PR #1, opened 2015-07-13
Create scheduled-message-works component for scheduling future delivery of RabbitMQ messages

dcohen:
I think we've managed to do everything in <a href="https://www.pivotaltracker.com/n/projects/930636/stories/96010356">this card</a> (and added some functionality to empty out S3 buckets as well) - please take a look and let us know!


ericnormand:
Pinging @ericnormand to bring me in the loop.

ericnormand:
Ping @cap10morgan can we get a review, please?

resources/config.edn
```diff
@@ -1,7 +1,15 @@
 {:rabbitmq {:connection {:host #resource-config/env "RABBITMQ_PORT_5672_TCP_ADDR"
                          :port #resource-config/edn #resource-config/env "RABBITMQ_PORT_5672_TCP_PORT"}
-            :queues {"scheduled-message-works.ok" {:exclusive false :durable true :auto-delete false}}}
+            :queues {"scheduled-message-works.ok"
+                     {:exclusive false :durable true :auto-delete false}
+
+                     "scheduled-message-works.message.schedule"
+                     {:exclusive false :durable true :auto-delete false}}}
  :datomic {:uri "datomic:mem://scheduled-message-works"
```
cap10morgan: This should be an env var.

resources/config.edn
```diff
@@ -1,7 +1,15 @@
...
            :partition :scheduled-message-works
            :intitialize true
```
cap10morgan: This should also be an env var.

src/scheduled_message_works/queue.clj
```diff
@@ -26,13 +26,21 @@
...
       (lq/declare ok-ch
                   "scheduled-message-works.ok"
                   (config :rabbitmq :queues "scheduled-message-works.ok"))
       (k/responder ok-ch "scheduled-message-works.ok" handlers/ok)
+
+      (lq/declare ok-ch
+                  "scheduled-message-works.ok"
+                  (config :rabbitmq :queues "scheduled-message-works.message.schedule"))
+      (k/responder ok-ch "scheduled-message-works.message.schedule"
+                   (handlers/handle-errors handlers/schedule-message))
```
cap10morgan: Looks like there are still some instances of `ok` in here that should be `schedule`. Channels and one queue name.

test/scheduled_message_works/handlers_test.clj
```diff
@@ -1,6 +1,42 @@
...
+            [datomic-toolbox :as db]
+            [turbovote.resource-config :refer [config]]
+            [clojure.test :refer :all]))
+
+(use-fixtures :each
+  (fn [test]
+    (d/delete-database (config :datomic :uri))
+    (db/initialize)
+    (db/run-migrations)
+    (test)))
```
cap10morgan: Is it too much to bite off to start doing this the Right Way™ here (i.e. creating a randomly-named Datomic DB each time so we aren't trying to delete and re-use the same test database over and over)?

test/scheduled_message_works/handlers_test.clj
```diff
@@ -1,6 +1,42 @@
...
+        ready-messages (message/get-ready-messages (db/db) (java.util.Date.))
+        rehydrated-messages (set (map message/rehydrate ready-messages))]
+    (is (= (count rehydrated-messages) 1))
+    (is (= :ok (:status result)))
+    (is (contains? rehydrated-messages (:message result)))
+    (is (thrown? Throwable (schedule-message {})))
+    (let [handler (handle-errors schedule-message)
+          response (handler {})]
+      (is (= :error (:status response)))
+      (is (:error response)))))
```
cap10morgan: A few `(testing ...)` forms in here would make it clearer which aspects of scheduling messages we're intending to test.

test/scheduled_message_works/handlers_test.clj
ericnormand: Do you know how to test this without modifying our config library?

cap10morgan: I think we'll need to modify datomic-toolbox, at least. Ideally (IMO) it would be decoupled from resource-config altogether.

But that's also a good indicator that yes, this is too much to bite off here.

cap10morgan:
LGTM

PR #2, opened 2015-07-16
Add identifiers upsert logic

ericnormand:
The purpose of this task was to make scheduling a message an upsert operation. When scheduling a message, a service will attach key/value pairs (identifiers) that it can use later as a handle into that message. A component might want to change or delete a message that has been scheduled before it goes out.

Each message does have an id, but it's arbitrary so the component will have to store and track the id to refer to the message later. The set of identifiers for a given message should uniquely identify the message, but should also be computable from data that the requesting service has. For instance, if I want to schedule a email for a given user and a given election and message type (template), those pieces of information identify the message uniquely. If something changes about the user (say, email address), I can simply issue another upsert for the updated message, and the identifiers should still match. The new message will replace the old one.

This component contains that basic logic: if the two sets of k/v pairs are equal, we go through an update path instead of insert. That means that the burden of wise and measured choice of the keys and values is pushed off as a burden on the requesting services. Those services know best the semantics of each message.

There are some limitations that we chose to impose on the identifiers. It cannot be empty. And the values must not be collections. Values are serialized and stored as edn strings and indexed for fast comparison.

Unfortunately, this brings up the notion of equality of collections vs equality of the string serializations of those collections. There are two problems. Clojure considers lists and vectors equal if they have the same elements in the same order. But these two values do not serialize to the same string. They are equal but not string equal. The second problem is that unordered collections (sets and maps) can be equal but print out their items in different orders, and so the strings are not equal.

We chose to limit the available identifier value types to those that are not collections to avoid these problems. other solutions exist, but we have yet to encounter a need for collection values.

The datomic query for doing the comparison is a little hairy. It does as much matching as it can by building up a large query with all of the key values of the new message (the known identifiers). If they all match, we will return a pair of (entity id, key) for all entities that match, with all of their keys.

We then have to group these back together by entity id. Now we check that all keys exist in the new message's identifiers. So we have both

1. All keys from A exist and match in B.
2. All keys from B exist in A.

That is equivalent to map equality.

[Pivotal Card](https://www.pivotaltracker.com/story/show/96011758)


ericnormand:
Test pass locally with

`lein run`

and `lein run :s3`

ericnormand:
Assigning to the inimitable @7maples

src/scheduled_message_works/entities/message.clj
```diff
@@ -55,10 +59,35 @@
                  :where [[?e :scheduled-message/id ?id]]}
                db id))
 
+(defn find-by-identifiers [db identifiers]
```
7maples: Can we add a comment explaining the basics of what this does and why, distilling the most important pieces from the PR description?

src/scheduled_message_works/entities/message.clj
```diff
@@ -68,6 +97,14 @@
...
     (entity->message (find-by-id (:db-after tx-result) id))))
 
+(defn update-message [existing-message-id new-message]
+  (s/validate schedule-message-schema new-message)
+  (let [existing-message (d/entity (db/db) existing-message-id)
+        tx-data (->tx-data existing-message-id (assoc new-message :id (:scheduled-message/id existing-message)))
+        tx-result @(db/transact tx-data)]
+    (entity->message (find-by-id (:db-after tx-result) (:scheduled-message/id existing-message)))))
+
+
```
7maples: extremely minor but if you're in here, there's an extra blank line

7maples:
LGTM

PR #3, opened 2015-07-23
[TH] adds delete scheduled-message functionality

tank157:
See the README for info on how to use. In addition
to being able to delete one or more messages by
identifiers, you can also delete individual
messages by their UUID.

[Pivotal Card](https://www.pivotaltracker.com/story/show/96013742)

Assigning to @ericnormand as he has good knowledge of this area already.


ericnormand:
This looks good.

README.md
```diff
@@ -87,6 +89,34 @@ or more do not match, then it is inserted as a new, unique message.
...
 
+### Deleting Messages
+
+To cancel a scheduled message, you can send a delete message in one of two formats, both sent
+to `scheduled-message-works.message.delete`.
+
+#### Deleting with Identifiers
+
+The format is the same as scheduling, but only contains the `:identifiers` key. You can delete multiple messages this way, and the response will include a `:count` of how many messages were deleted.
+
```
ericnormand: I see that matching by identifiers works differently in delete. Namely, in scheduling a message, the upsert is an update if the identifier maps are equal. However, in delete, only the given identifiers have to match. What do you say about highlighting this difference in the readme?

src/scheduled_message_works/entities/message.clj
```diff
@@ -59,6 +59,32 @@
...
+          (build-identifiers-query identifiers)
+
+          eks
+          (d/q {:find '[?e]
+                :in '[$]
+                :where all-identifiers-match}
+               db)]
+      (map first eks))))
+
 (defn find-by-identifiers
```
ericnormand: Now that find-by-identifiers and search-by-identifiers both exist, is it possible to find names that indicate how they are different? search and find have very similar meanings.

tank157: My understanding on this, and I could be wrong, is that we've been using `find` when we expect at most 1 result, and `search` when we could get multiple. So I tried to keep that going here.

ericnormand:
Also, it looks like you've changed the notification-type to be a keyword. Would you mind changing that in the docs as well? https://torchpad.com/workspace/wikis/democracyworks/pages/research/Notifications+Scheduling

src/scheduled_message_works/entities/message.clj
tank157: And now at the risk of pointing to something _I just created_, after conferring with Chris that this is indeed our unwritten standard, I've now started a place in torchpad for Code Conventions/Name Conventions and documented this there. https://torchpad.com/workspace/wikis/democracyworks/pages/Coding+Conventions/Naming+Conventions

ericnormand: Ok, that sounds good.

ericnormand:
LGTM

ericnormand:
Checked torchpad and the docs are there.

PR #4, opened 2015-07-29
Trigger sends

tank157:
[Pivotal Card](https://www.pivotaltracker.com/story/show/96014532)

Adds a scheduler to trigger the sending of scheduled messages, using the chime library and a periodic-seq. I calculate the next evenly divisible minute based on the configurable granularity, so that works best with granularities that divide evenly into the hour (1, 2, 3, 4, 5, 6, 10, 15, 20, 30, 60). I suppose it could still work otherwise, but then your send time is going to differ from hour to hour. This then becomes the start of the periodic-seq and it repeats every granularity number of minutes.

I had to fix some broken tests in the delete that mainly came down to a lazy-seq not being evaluated.

Assigning to @tie-rack since I'm sure he can provide some great feedback.


src/scheduled_message_works/entities/message.clj
```diff
@@ -78,12 +79,15 @@
...
           (build-identifiers-query identifiers)
 
-          eks
-          (d/q {:find '[?e]
-                :in '[$]
-                :where all-identifiers-match}
-               db)]
-      (map first eks))))
+          query {:find '[?e]
+                 :in '[$]
```
tie-rack: The database as `$` is implicit when there are no other arguments to the query, so this shouldn't be necessary.

src/scheduled_message_works/handlers.clj
```diff
@@ -71,15 +71,16 @@
...
   {:status :ok
    :count 1})
 
 (defn delete-message-by-identifiers
   [identifiers]
-  (let [messages (message/search-by-identifiers (db/db) identifiers)]
-    (dorun (map message/delete-message messages))
+  (let [messages (message/search-by-identifiers (db/db) identifiers)
+        uuids (map :scheduled-message/id messages)]
+    (dorun (map message/delete-message uuids))
```
tie-rack: I think a `doseq` is more conventional here than `(dorun (map...` since we're not using the value returned by `map`.

src/scheduled_message_works/scheduler.clj
```diff
@@ -0,0 +1,83 @@
+(ns scheduled-message-works.scheduler
+  (:require [scheduled-message-works.channels :as channels]
+            [scheduled-message-works.entities.message :as message]
+            [kehaar.core :as k]
```
tie-rack: Kehaar is not used in this ns.

src/scheduled_message_works/handlers.clj
tank157: But nor would I be doing anything with the binding expressions that doseq provides.

src/scheduled_message_works/scheduler.clj
```diff
@@ -0,0 +1,83 @@
...
+  and sends them."
+  [sent-at-before]
+  (log/debug "sending scheduled messages from before" sent-at-before)
+  (if-let [ready-messages (message/get-ready-messages
+                           (db/db)
+                           (coerce/to-date sent-at-before))]
+    (->> ready-messages
+         (map message/rehydrate)
+         (map send-message)
+         dorun)))
```
tie-rack: Since this is a fn used purely for side-effects and the result of `(map send-message)` isn't used, this might also be more straightforward as a `doseq`:

```clojure
(doseq [message (map message/rehydrate ready-messages)]
  (send-message message))
```

src/scheduled_message_works/handlers.clj
tank157: Though I suppose I could rewrite it to use them, that just feels weird when `dorun` already seems suited to this purpose.

tank157: Well, with an obscure warning about chunked sequences, looks like the recommendation is to avoid the `dorun` style of implementation: http://w3facility.org/question/in-clojure-is-there-a-doseq-like-function-that-receives-a-function-instead-of-body/

Will `dorun` then be deprecated and removed? It seems made for this exact use case.

src/scheduled_message_works/scheduler.clj
```diff
@@ -0,0 +1,83 @@
...
+(defn next-start-time
+  "Given a start time and granularity, figures out
+  the next time where the minute is evenly divisible
+  by that granularity (in minutes). So a granularity
+  of 10 and a current time of 6:03:49am, it will
+  return 6:10:00am."
+  [start granularity]
+  (let [minute-of-hour (time/minute start)
+        upcoming (iterate inc minute-of-hour)
+        next-match (->> upcoming (filter #(= 0 (mod % granularity))) first)
```
tie-rack: Instead of filtering over an infinite sequence, isn't this directly computable?

```clojure
(let [minute-of-hour (time/minute start)
      next-match (+ minute-of-hour (- granularity (mod minute-of-hour granularity)))
      ...
```

src/scheduled_message_works/scheduler.clj
```diff
@@ -0,0 +1,83 @@
...
+  "Given a start time and granularity, figures out
+  the next time where the minute is evenly divisible
+  by that granularity (in minutes). So a granularity
+  of 10 and a current time of 6:03:49am, it will
+  return 6:10:00am."
+  [start granularity]
+  (let [minute-of-hour (time/minute start)
+        upcoming (iterate inc minute-of-hour)
+        next-match (->> upcoming (filter #(= 0 (mod % granularity))) first) 
+        rounded-start (-> start (.hourOfDay) (.roundFloorCopy))]
```
tie-rack: No need for those parens: `(-> start .hourOfDay .roundFloorCopy)`

src/scheduled_message_works/scheduler.clj
tank157: Sure, but my way is more fun.

src/scheduled_message_works/handlers.clj
tie-rack: `dorun` is suited to the purpose of having a lazy sequence (that you want) whose computation includes side effects. But since we don't want the sequence, why even produce it in the first place? `doseq` is pretty explicit in its meaning: do this side-effecty thing to everything in this seq.

I'm not sure what you mean by "nor would I be doing anything with the binding expressions that doseq provides", since you'd be using whatever you bind for each member of the seq in a call to `message/delete-message`.

tank157: Well, at the time I was envisioning a straight up swap of dorun and doseq, without thinking about how it would need to be rewritten.

But in all honesty, had I not found higher up (ie Cognitect) folks saying to avoid dorun with lazy-seqs, I wouldn't really see the point in changing to doseq. The dorun map reads just fine, and avoids noisy temporary bindings. Had the map been more complicated, sure, but it was pretty straightforward as is.

tank157:
@tie-rack ok, I think this is ready for one last look. I had to change the `next-start-time` implementation you suggested to allow it to basically return now if it lined up properly. Otherwise the tests got cranky.

tie-rack:
Looks good!

PR #5, opened 2015-07-30
[TH] add read and search queues

tank157:
Not strictly called for yet, but nice to have
to demo the delete fns. So there isn't a Pivotal
Card for this.

Assigning to @cap10morgan


src/scheduled_message_works/handlers.clj
```diff
@@ -89,3 +94,19 @@
...
     (delete-message-by-identifiers (:identifiers identifying-data))
     (delete-message-by-uuid (:id identifying-data))))
+
+(defn read-message
+  [message-data]
+  (s/validate uuid-schema message-data)
+  (if-let [message (message/find-by-id (db/db) (:id message-data))]
+    {:status :ok
+     :message (message/entity->message message)}
+    (throw (IllegalArgumentException. "Not found"))))
```
cap10morgan: Should we include the id that it failed to find in the error message?

tank157: We haven't been doing so anywhere else we're doing the same thing yet. It's a direct response type of call, so the caller should have the id already. But I could be convinced we need to adjust our informal standard around this if you think it's still necessary.

cap10morgan: I'm just thinking about when we're tailing logs in production later. Getting the id there is really helpful.

cap10morgan: But I see that right now handle-errors isn't logging these. So fine, we can fix that later at a different level (something like handle-errors feels destined for a library of some sort).

cap10morgan:
LGTM

cap10morgan:
Oops, sorry, one more question: Should we add anything to the README about these changes?

tank157:
Ok, I've updated the README and added the id to the Not Found case. One more quick look @cap10morgan ?

cap10morgan:
LGTM

PR #7, opened 2015-08-05
Refactor upsert into entities

ericnormand:
This refactors the upsert functionality into the entities namespace.


cap10morgan:
@tie-rack Should we merge this one (after scaling work lands)? Other than the private fns (I'd be in favor of making those public again), it seems like an OK change to me.

tie-rack:
It looks like it would behave the same... but this seems to need to be written with `swap-tx!`. This is a race condition waiting to happen.

cap10morgan:
@tie-rack Think we should close this one unmerged for now? And then maybe leave the branch around in case it's still a good basis for a future improvement?

tie-rack:
@cap10morgan Sounds good. Closing.

PR #9, opened 2015-08-06
make file-storage-location settable via ENV VAR

cap10morgan:
FILE_STORAGE_LOCATION

Assigning to you, Eric (even though we did this together and it's very simple) so that you can coordinate the production config & deployment issues.


ericnormand:
Can you set this env var in the service template?

ericnormand:
Well, assuming the merge goes well, LGTM.

ericnormand:
LGTM

PR #11, opened 2015-10-29
Add vote tomorrow update

cap10morgan:
This is a fairly urgent PR needed for vote-tomorrow notifications to go out on Monday (I thought it had already been put up and merged).


ericnormand:
LGTM

cap10morgan:
Just so we write it down somewhere other than Slack: Eric and I agreed that it's very unfortunate that you have to add so much supporting code to scheduled-message-works for something that it should not have to care about (i.e. which queue a scheduled message wants to be sent on when it's :send-after time arrives).

We're planning to write up a research card about eliminating the need for this.

PR #14, opened 2016-03-07
Fix race condition in e.message/update-message

cap10morgan:
We were assuming we could get a message entity :db/id argumeent from the
latest db value when it may have been deleted before we got it. And in
fact, that's what we saw in the production logs today.

This at least makes it stop assuming that and makes
`find-by-identifiers` return a Datomic entity like most other
`find-by-*` fns do.

We'd like to get this deployed soon if it looks OK since it is causing errors in production.

Also since there's no card (maybe we should create one?) I'll note that @tank157 and I paired on this.


src/scheduled_message_works/entities/message.clj
```diff
@@ -114,13 +114,15 @@
...
-            entity-keys))))
+          (set (keys identifiers))
+
+          ;; we return the entity-id of some message where all keys are a
+          ;; subset of the `identifers` keys
+          db-id (some (fn [[e eks]]
+                        (when (every? id-keys (map second eks))
+                          e))
+                      entity-keys)]
+      (d/entity db db-id))))
```
tie-rack: The docstring says "Returns the message entity if one is found", but there's no conditional on whether or not one is found. It's always trying to return an entity. What happens if one isn't found? Or does Datomic return some kind of null entity if the `db-id` is nil?

cap10morgan: It returns `nil`. I can add that to the docstring.

src/scheduled_message_works/entities/message.clj
```diff
@@ -114,13 +115,15 @@
...
-      ;; we return the entity-id of some message where all keys are a
-      ;; subset of the `identifers` keys
-      (some (fn [[e eks]]
-              (when (every? id-keys (map second eks))
-                e))
-            entity-keys))))
+          (set (keys identifiers))
+
+          ;; we return the entity-id of some message where all keys are a
+          ;; subset of the `identifers` keys
```
tie-rack: This comment seem out of date now, both in the wrong place (we are not at the return of the function anymore) and incorrect (we now return an entity, not an entity id).

src/scheduled_message_works/entities/message.clj
tie-rack: Interesting. I could have seen that go either way (returning nil or throwing a NPE).

tie-rack:
Minor nitpick about the comment notwithstanding, this LGTM.

PR #19, opened 2016-07-11
Return rehydrated messages from read & search

cap10morgan:
Not very useful otherwise.

Doesn't look like anything else was using these, but I'm about to in my election-mail-subscription event handlers that update notifications w/ new election mail subscription statuses.

I need the full message contents (what `rehydrate` provides) in order for that to work.

Not sure what Pivotal card to link to here. The card ostensibly calling for this work is already accepted, which I don't agree with. And then there are various bug cards which were created prematurely and none of which really captures what's going on here. So since we're not standing on formality in Pivotal, I guess I won't do so here either?


ddellacosta:
LGTM

PR #21, opened 2016-09-07
New relic and kehaar

danielglauser:
Adds NewRelic tracing. Does not upgrade Kehaar since there are major changes between 0.4.0 and 0.6.0


src/scheduled_message_works/file_storage.clj
```diff
@@ -3,49 +3,50 @@
...
-           [aws.sdk.s3 :as s3]))
+           [aws.sdk.s3 :as s3]
+           [com.climate.newrelic.trace :refer [defn-traced]]))
 
 (def default-charset "UTF-8")
 
 (defonce temp-directory (.toFile (Files/createTempDirectory "messages" (into-array FileAttribute []))))
 
-(defn store-file-temp-dir [location content]
+(defn-traced store-file-temp-dir [location content]
```
tank157: this is only used in local mode, don't really want tracing on here.

src/scheduled_message_works/file_storage.clj
```diff
@@ -3,49 +3,50 @@
...
                  {:content-type "text/plain; charset=UTF-8"}))
 
-(defn store-file [location content]
+(defn-traced store-file [location content]
   (case (config [:file-storage-location] :file)
     :file (store-file-temp-dir location content)
     :s3   (store-file-s3       location content)))
 
-(defn read-file-temp-dir [location]
+(defn-traced read-file-temp-dir [location]
```
tank157: I think the same is true here.

PR #22, opened 2016-09-08
Allow disabling the scheduler via config param

cap10morgan:
This will allow us to scale scheduling but not sending ("scheduler"
is perhaps a bad name for that part of the system; it should probably be
renamed "sender").


docker-compose.yml
```diff
@@ -22,9 +23,11 @@ wildfly:
     ADMIN_USERNAME: admin
     ADMIN_PASSWORD: admin
 datomic:
-  image: quay.io/democracyworks/datomic-tx:0.9.5327
+  image: quay.io/democracyworks/datomic-tx:0.9.5394
+  environment:
+    ALT_HOST: datomic
 rabbitmq:
-  image: rabbitmq:3.5.3-management
+  image: rabbitmq:3.6.2-management
```
danielglauser: @cap10morgan Are we sure this is okay with the older version of Kehaar (0.4.0)?

cap10morgan: Yes

danielglauser:
LGTM

danielglauser:
LEBTM (Looks Even Better To Me)

danielglauser:
LGTM

PR #23, opened 2016-09-14
Add tracing to the entity ns functions

danielglauser:


danielglauser:
Troy said it was good, I swear

PR #24, opened 2016-09-14
Adds an index for identifiers

danielglauser:


danielglauser:
Chris said this was okay, yolo.

PR #25, opened 2016-09-14
adds an identifier hash for use with find-by-identifiers

tank157:
Will need some manual migrations after it's deployed


src/scheduled_message_works/entities/message.clj
```diff
@@ -164,25 +196,48 @@
...
 (defn-traced delete-message [id]
   (log/debug "deleting scheduled message with id" id)
-  @(db/transact [[:db.fn/retractEntity [:scheduled-message/id id]]]))
+  (when-let [message (find-by-id (db/db) id)]
+    (let [identifier-ids (map :db/id (:scheduled-message/identifier message))
+          all-ids (conj identifier-ids (:db/id message))
+          retract-tx-data (->> all-ids
+                               (map ->retract-tx-data)
+                               vec)]
+      @(db/transact retract-tx-data))))
```
cap10morgan: After reading the docs, it looks like it would be safe to just make identifiers components as that bug card calls for. That would also make this simpler. So I would vote for starting there instead.

danielglauser:
Tested locally in Krakenstein.

PR #26, opened 2016-09-15
Remove the old `find-by-identifiers`

danielglauser:
Stop calling the old `find-by-identifiers` which could take a long time.


src/scheduled_message_works/entities/message.clj
```diff
@@ -128,7 +128,7 @@
   [identifiers]
   (digest/sha-1 (pr-str (into (sorted-map) identifiers))))
 
-(defn-traced find-by-identifiers-hash
+(defn-traced find-by-identifiers
   "Hashes the identifiers and searched for that specific value
```
tank157: I know this isn't in the changes, but can you fix my typo on 'searched'? I think it should be "searches".

tank157: Otherwise LGTM if it passes tests.

PR #30, opened 2016-09-24
Fix handling of default configuration values

gws:
`turbovote.resource-config/config` determines lookup success based on whether keys exist, and so will not return default values when the key path exists but the values are nil.

Without this or a similar change, if the environment variable `SCHEDULED_MESSAGE_WORKS_UNRESERVE_OLDER_THAN_IN_MINUTES` is not set (e.g. in a development environment), a NPE will occur in `scheduled-message-works.send-messages/unreserve-messages` when trying to calculate `nil` minutes ago.


cap10morgan:
@tank157 Please merge at your convenience.

cap10morgan:
@tank157 Any reason not to merge this?

tank157:
Sorry, I missed the request for me to merge it (which is generally not our workflow, but I guess I understand why). Let me take another quick look and I can see about merging it.

PR #37, opened 2016-10-08
Add retries to all Datomic transactions and queries

danielglauser:


src/scheduled_message_works/entities/message.clj
```diff
@@ -167,15 +186,18 @@
...
-  (let [identifiers-hash (->identifiers-hash identifiers)
-        entities
-        (d/q '{:find [[?e ...]]
-               :in [$ ?identifiers-hash]
-               :where [[?e :message-identifiers/combined-hash ?identifiers-hash]]}
-             db identifiers-hash)]
-    (some->> entities
-             first
-             (d/entity db))))
+  (try-try-again
```
ddellacosta: Why do we want this here if we already call `try-try-again` in the functions calling this (`update-message` and `schedule-message`, looks like)?

src/scheduled_message_works/entities/message.clj
```diff
@@ -87,18 +94,24 @@
 (defn-traced unreserve-messages
   "Releases reserved messages that were reserved before as-of."
   [as-of]
-  (let [{:keys [tx-data]}
-        @(db/transact [[:scheduled-message/unreserve-messages as-of]])
-        num-unreserved (- (count tx-data) 1)]
-    (when (> num-unreserved 1)
-      (log/warn "Found" (- (count tx-data) 1) "stagnant reserved messages"))
-    num-unreserved))
+  (try-try-again
```
ddellacosta: Here again we are calling `try-try-again` in the (a?) function that calls this (`scheduled-message-works.send-messages/remove-message`), and same with `reserve-messages` here and in the `send-messages` ns--do we really need this redundancy?  Seems like it could theoretically even cause problems, although I have to admit I'm not sure how yet.  Regardless, I wonder if it wouldn't be better to add this sparsely for now, and then add more promiscuously if it seemed like the right move.

cap10morgan:
@ddellacosta & @danielglauser Where is this at? Can we move it forward or close it?

ddellacosta:
@cap10morgan Reviewing my comments just now, I _think_ I'm still waiting to hear back from Daniel on a few of these-- @danielglauser can you confirm?

src/scheduled_message_works/entities/message.clj
danielglauser: I removed the call to `try-try-again` in `send-messages`.

src/scheduled_message_works/entities/message.clj
danielglauser: Since I removed all the calls in the handler `ns` I believe I've removed all the redundancy. This now only retries calls to Datomic.

danielglauser:
@ddellacosta I think my last commit addressed all of your comments, please let me know if you think otherwise.

danielglauser:
I've spent a while trying to test this in Krakenstein. Unfortunately it's proving rather difficult to cause the retries to happen, however I was able to confirm that the happy path work just fine. Even in the case where the retries don't work (unlikely because this library has been really well tested in production for years) it seems that we won't be any worse off.

src/scheduled_message_works/entities/message.clj
ddellacosta: Sorry looks like I was confused by github's formatting, and thought this was the body of `->identifiers-hash`.

danielglauser:
@tank157 I'd either like to get this deployed or close the PR. Do you want this work in `scheduled-message-works`as is or should I create a story for adding retires to [datomic-toolbox](https://github.com/democracyworks/datomic-toolbox)?

danielglauser:
@tank157 Please have another look.

PR #40, opened 2016-10-25
add tracing to scheduler processes

tank157:
I decided to take the unusual (for me) tack of tracing everything in this namespace to get both the overall picture of time spent, and the micro-picture when just a single message is in flight. The overall picture won't be that useful right now, as we don't know how many messages are in each run, though the individual send-message traces during the same period should give us an average. Anyway, the overall picture will get clearer when we're sending all the messages in big batches.


danielglauser:
LGTM dawg

PR #42, opened 2016-12-27
Upgrade to kehaar 0.8.1

tie-rack:
Configured!


resources/config.edn
```diff
@@ -1,33 +1,45 @@
...
-
-                     "election-mail-works.mailing.vbm-update"
-                     {:exclusive false :durable true :auto-delete false}
-
-                     "ballot-scout.ballot.delivery-update"
-                     {:exclusive false :durable true :auto-delete false}
-
-                     "mail-works.mailing.send"
-                     {:exclusive false :durable true :auto-delete false}}}
+                     :outgoing-events
```
tank157: I'm not as up-to-speed on the 0.8 conventions, so these events are outgoing because they go out from other services? I sort of see them as incoming events myself.

tie-rack: They are "outgoing events" because they're replacing `wire-up/outgoing-events-channel` declarations from queue.clj. These are messages going from the process out to rabbit.

tank157: Oh, and they're going as events because they are fire-and-forget?

tie-rack: Huh... they _should have been_ using the fire-and-forget mechanisms already if that's what they are... weird.

So, yeah, this is a translation of _what was there_, not (perhaps) how it should be.

tank157: I do believe fire-and-forget is the right thing for these to do, scheduled-message-works is just passing the messages along, can't really react to anything much. So yeah, I bet they got shoe-horned in as events before we had f-n-f.

PR #44, opened 2017-01-18
Add Datomic obj cache max setting

cap10morgan:
Just like in user-works


cap10morgan:
@tank157 & @ddellacosta Should we merge this now?

ddellacosta:
@cap10morgan  Sorry for the slow response Wes--fine from my perspective.

PR #49, opened 2017-04-28
Configurable run message forwarder/scheduler

cap10morgan:
These changes should allow us to run this in the TV admin cluster.

I've already set the relevant Consul values.

No card because it was a Friday project.

Nitpicky aside: I don't like calling this entire subsystem the "scheduler." Scheduling messages is a separate feature of scheduled-message-works, and this name is highly confusing with that. This is more of a message forwarder (i.e. this periodically checks for all scheduled messages whose time-to-forward has come due, and this then does the forwarding). I think the name comes from the fact that it uses an internal feature in s-m-w that is distinct from the external message scheduling but also named `scheduler` (that's the namespace). This internal "scheduler" provides the cron-like "run this fn periodically" functionality. But even there I would prefer a cleaner distinction between the _what_ (message forwarding at regular intervals) and the _how_ (internal scheduler feature). A refactor to clear all this up seemed bigger than was appropriate for now, though.


src/scheduled_message_works/core.clj
```diff
@@ -10,20 +10,54 @@
...
-        (or (config [:scheduler :batch-size] 100) 100)
-        unreserve-older-than-in-minutes
-        (or (config [:scheduler :unreserve-older-than-in-minutes] 30) 30)
-        periodic-fn (partial send-messages/reclaim-and-send
-                             batch-size unreserve-older-than-in-minutes
-                             channels/queue->channel)
-        scheduler-ch (scheduler/start granularity periodic-fn)]
+  (let [rabbit-resources (queue/initialize (rabbitmq-config))
+        scheduler-config {:granularity (config [:scheduler :granularity])
+                          :batch-size (or (config [:scheduler :batch-size] 100) 100)
```
ddellacosta: Why do we need the `or` here and immediately below if `config` provides us with the ability to specify a default?

tie-rack: I can tell you why! `(config [:scheduler :batch-size] 100)` will return `nil` if the key path exists but has `nil` as its value. And that's what it would do if [the env var that populates that key](https://github.com/democracyworks/scheduled-message-works/blob/957c541dbd953ffde1cb74eba003a9d7bb39fe2d/resources/config.edn#L53-L55) is missing.

cap10morgan: blam

ddellacosta: That is terrible. But thanks for the explanation.

PR #50, opened 2017-08-11
Identifier overhaul

tie-rack:
Instead of the cardinality many key/value identifier component entities, identifiers are attributes directly on message entities. Previously unseen identifiers are added to the database schema on the fly.

Searching for messages by identifiers is now many many times faster. The danger is that new identifiers could keep coming, creating a new index each time. However, in practice there are very few identifiers in use.

To migrate messages from the old identifier system to the new one, you can run:
```
lein run -m scheduled-message-works.identifiers-translation
```

This will create the new identifier attributes for all existing messages, add datoms with those new attributes for existing messages, and retract the old identifiers.

When deployed, system downtime to migrate messages may not be necessary if we avoid doing those things that trigger searching/finding by identifiers between deploying and running the identifiers translator. I would imagine "those things" to include (but not be limited to) election updates and template updates.


src/scheduled_message_works/entities/message.clj
```diff
@@ -112,68 +125,56 @@
...
-
-(defn-traced search-by-identifiers*
+  (->> identifiers
+       sort-identifiers
+       (map (fn [[k v]]
+              ['?e (key->identifier-attribute k) (pr-str v)]))))
+
+(defn all-identifier-keys? [db identifiers]
+  (= (count identifiers)
+     (ffirst (datomic.api/q '{:find [(count ?e)]
```
gws: You could drop the `ffirst` wrapper and change the `:find` value to `[(count ?e) .]`.

tie-rack: Oh thank you!

resources/schemas/2017-08-08-save-identifier-fn.edn
```diff
@@ -0,0 +1,29 @@
...
+                               :where [[_ :db/ident ?identifier]]}
+                             db
+                             identifiers))
+         to-create (set/difference
+                    identifiers
+                    known-identifiers)]
+     (when (seq to-create)
+       (map (fn [identifier]
+              {:db/id (datomic.api/tempid :db.part/db)
+               :db/ident identifier
```
zonotope: where do these new identifiers come from in practice? since all idents have to be stored in memory, i'm a little worried that a bug in another service's code could write brand new idents in each request, and all those idents get added on the fly, which would blow us up eventually.

tank157: New identifiers come from whatever service is scheduling a message to come later. Primarily that's `election-notification-works` and `election-mail-works` currently (sometimes by way of `notification-works`, but just as a pass-through).

There is the possibility of unbounded growth here, yeah, if someone keeps passing unique strings in as the identifiers. In practice it hasn't happened in the old system either, but the danger is lessened since they are just values at `:message-identifier/key`, and DBs are great at storing lots of values. Is there some way we could keep the flexibility but monitor the growth? Maybe we could broadcast an event every time a new identifier comes in and keep an eye on a queue listening for those events?

tie-rack: Currently `election-notification-works`, `mail-works`, and `notification-works` send messages that could result in new identifiers. And, yes, a bug in another system could do that. The implicit bet that I'm willing to make explicit is that 1) we won't do that (generating fresh keys in maps for these requests would seem to require much more than fat fingering a name) and 2) the danger of that happening is much less real than the danger that `scheduled-message-works` will not be able to handle the volume of messages we're going to be seeing.

tank157: Also of concern here is that all identifiers are getting marshaled to Strings. I'm just trying to think if there's a good reason to try and maintain their types for indexing purposes. A lot of identifiers will be UUIDs generated by `d/squuid`, and I wonder if performance would be better with them as UUIDs? It might be negligible.

src/scheduled_message_works/entities/message.clj
```diff
@@ -32,25 +37,33 @@
...
-  [(:message-identifier/key entity)
-   (-> entity
-       :message-identifier/value
-       edn/read-string)])
+  (into [(tx/dissoc-optional-nil-values
+          {:scheduled-message/send-to send-to
+           :scheduled-message/send-after send-after
+           :scheduled-message/send-before send-before
+           :scheduled-message/content-hash content-hash
+           :message-identifiers/combined-hash identifiers-hash
```
tank157: Do we need to continue using the `combined-hash` in this new system? It was implemented as a really stop-gap measure to help use get through the 2016 Presidential election and I wouldn't mind seeing it go away if the new system is better.

resources/schemas/2017-08-08-save-identifier-fn.edn
gws: Since the potential issue here is just about holding `known-identifiers` in memory on the transactor, could we reformulate the query to return only the `identifiers` for which there is not a known ident? That's ultimately what line 20+ want.

Example:

```clojure
'{:find [[?identifier ...]]
  :in [$ [?identifier ...]]
  :where [(not [_ :db/ident ?identifier])]}
```

Unbounded growth is literally a selling point of Datomic, and I'm not too worried about adding new data to a database :smile:

gws: Folks, I'm gonna leave this here, but I thought we were talking about unbounded memory usage on the transactor, because of:

>since all idents have to be stored in memory

and not a large number of indexed attributes.

So, disregard that.

zonotope: yeah, i'm also not concerned with adding data to the database, and i wasn't really thinking of `known-identifiers`, but what you brought up @gws could also potentially be a problem.

i am a little concerned with modifying the schema based on external input and adding new idents to datomic based on that input because [the set of all a datomic instance's idents has to fit in memory both on the transactor and peers](http://docs.datomic.com/identity.html#idents). if for some reason we get a service that has some bug that makes a new, unique database schema attribute for each request along with that attribute's corresponding ident, we'd exhaust the ram on the transactor pretty quick. 

it wouldn't necessarily take a perfect storm kind of bug for this architecture to present a problem. something simple as `(map (fn ... [v k]))` instead of `(map (fn ... [k v]))`, where `k` is the attribute name and `v` is a unique value. 

in any case, the alternatives i can think of could (probably would) fall over during normal operation on a heavy day anyways, as @tie-rack suggested. if no one else is worried about unbounded idents, then i won't.

gws: Well, the other piece that I missed at first because I didn't check "below the fold" was that I saw references to `:db/ident` and adding more of those, as [enums](http://docs.datomic.com/schema.html#enums) isn't a problem. In fact, this seems like a really good use case for them.

The only drawback seems to be that they need to be keywords. You could construct them as:

`:message-identifier/keyword-safe-string`

tie-rack: > Also of concern here is that all identifiers are getting marshaled to Strings. I'm just trying to think if there's a good reason to try and maintain their types for indexing purposes. A lot of identifiers will be UUIDs generated by d/squuid, and I wonder if performance would be better with them as UUIDs? It might be negligible.

This problem stems from the same source as the unbounded identifier key problem (which also exists in current `master`, though it exhibits itself differently) The source of both of them is the fact that scheduled-message-works is _dumb_. It doesn't understand anything about the data—either what the values are or how they're keyed—by design.

As it stands, the contract scheduled-message-works upholds is: you can give me an arbitrary map and I'll let you search for messages via subsets of that map and be able to return the full thing. That's the contract I was maintaining in this PR.

If we wanted to change the contract to add some constraints, there are some further performance improvements we could gain. For example, we could specify that any key that ended in `-id` had to have UUID values so we could more effectively index those.

zonotope: @gws unbound enums in datomic database are also a problem because they're also just idents. having a previously specified finite set of enums is fine, but adding them on the fly based on user input could cause trouble.

gws: @zonotope Enums are just entities in Datomic, just like any other entity, but **not** like schema attributes. As a relational database analogy, enums are similar to already having a global single enum table with two columns: an integer primary key and a keyword (:db/id and :db/ident), with one **row** for each enum, and the rest of the database refers to the enum by the indexed primary key (with an index on the keyword part as well to support pretty queries). That's a completely different world from installing a new schema attribute for every new `identifier`, which is analogous to adding an indexed **column** to a table on the fly, which is what is being proposed in this PR right now.

tie-rack: > analogous to adding an indexed **column** to a table on the fly

And this is where the relational database analogy starts to break down. There's not a huge table that's getting a new column. The attribute being added has no intrinsic attachment to any set of data. There's not going to be a big set of `NULL`s generated for all existing rows of some table. In a relational database, adding indexed columns on the fly would be bonkers.

gws: @tie-rack it's true, and it's a very fair point. it's a much lighter weight operation in Datomic than in most RDBs. Some of them have sparse column support, though - I was angling less for the "this operation is gross and expensive" and more trying to illustrate that enums in Datomic are cheap and, IMO, a good conceptual fit.

src/scheduled_message_works/entities/message.clj
tie-rack: An argument for keeping it might be that a query on a single indexed attribute that's going to return a single value is probably going to be faster than a query against multiple indexed attributes, none of which would be enough to filter down to a single value.

resources/schemas/2017-08-08-save-identifier-fn.edn
zonotope: @gws i agree (roughly) with your analogy. the point i was trying to make is that the global single enum table (to continue the analogy) that you speak of is special and not like other database tables because it has to fit completely in the transactor's RAM, even if those enums aren't actively being accessed, while the other information and entities in the database can be stored on disk through datomic's storage back end if it isn't actively being accessed. from the datomic "Identity and Uniqueness" doc:
> Idents are designed to be extremely fast and always available. _All idents associated with a database are stored in memory in every Datomic transactor and peer._

the amount of ram available on the transactor will be much, much less than the amount of space available to the storage back end. unbound data being appended to the storage back end is fine and what the system is designed for. possibly unbound data appended to the transactor's memory is not. 

one or two extra enums of course won't cause a problem, but since we're just adding new enums (along with their corresponding idents) on the fly based on external input automatically, there is the potential for an external process that has a bug or a misunderstanding of the limitations or purpose of these identifiers to continuously create new and unique enums systematically with every request, then the RAM on the transactor will be exhausted eventually, and that will happen much faster than any storage back end will fill up.

this is not to say that we should not move forward with this approach, but if we do move forward with this approach, this is something we should be aware of.

tank157: Could we add some kind of "maximum number of unique identifiers" check? It would only have to get run when we're actually adding a new identifier. We've gone almost a year with fewer than 10 identifiers, so maybe we set it at some stupidly high number like 100 before we throw an error.

src/scheduled_message_works/entities/message.clj
tank157: Ok, so long as it's still in use somewhere, which it probably is in the exact match case.

resources/schemas/2017-08-08-save-identifier-fn.edn
gws: @zonotope Thank you! I missed that you were referencing that bit of the doc the first time around, and I was not taking that into consideration during this discussion. Sorry about that. I did understand the point you were making at the start, I just wasn't sure where the exhaustion was taking place, but I see it now.

src/scheduled_message_works/entities/message.clj
tie-rack: Correct. The `schedule-message` handler uses it to determine if a message is being created or updated.

resources/schemas/2017-08-08-save-identifier-fn.edn
tie-rack: Somehow limiting the number idents created is a laudable goal, but I'd like to do it in a way that doesn't feel totally arbitrary (like setting a maximum number of idents), allows clients to be flexible in how they choose to identify messages, and doesn't require scheduled-message-works to have prior knowledge of the idents.

How about something like this: when a service that uses scheduled-message-works starts up it can send a message to scheduled-message-works that says "here are the identifiers I'm going to use", and s-m-w will limit the idents created to the union of all such messages it gets over its lifetime. Attempts to schedule messages from services that include identifiers outside of that set are rejected.

tank157: Hmmmm, could be a good way to go about it. You could still have unbounded growth, if someone requested unique identifiers every time it started up, for instance, but the growth curve would be much slower.

zonotope: @tie-rack i think that could be a good compromise given the constraints you mentioned. at least we wouldn't be creating new idents on the fly with new requests, so the potential for a bug in another service to crash our transactor is a lot smaller.

tie-rack:
Alright, @gws, @tank157, and @zonotope. The number of attributes that can be created is now limited by a configurable value.

tank157:
Didn't feel like doing the thing where services broadcast the identifiers they planned to use on startup? Not that I blame you, it's a tidy solution in some respects, but then needs work in `election-notification-works` and `election-mail-works`. I'm down with this though.

PR #51, opened 2017-10-03
Use typical Docker CMD

cap10morgan:
In our other components, we let the cgroup-set heap size auto-determine the Datomic obj cache (among other things).

This was crashing in k8s dev due to insufficient RAM for the requested Datomic obj cache size.


cap10morgan:
Yeah, that would make sense. Though there was a time where we started adding this to several components before the JVM folks added the CGroup parameter we use now. So it could be a hold-over from that. It looks like we are limiting it to `2g` currently in tvprod (it would be much bigger if left to the 50% of available JVM RAM default). I'm not sure if we still need that or not.

If we do, I'm happy to re-work this so it better accommodates the present and the future.

PR #56, opened 2018-01-22
Log when Robert the Bruce retries

tie-rack:


src/scheduled_message_works/entities/message.clj
```diff
@@ -19,7 +19,12 @@
   the settings below the longest it will retry is for 4.89 minutes.
   That was determined by the summation of e^0 - e^5, assuming the
   decay is exponential."
-  {:decay :exponential :sleep 1000 :tries 5})
+  {:decay :exponential :sleep 1000 :tries 5
+   :error-hook (fn [error]
+                 (log/error "Robert Bruce retry" error)
```
cap10morgan: Can you `pr-str` that error? Or does it not help much w/ error objects?

PR #60, opened 2018-01-24
Pipelining of Datomic transactions and S3 file uploads

cap10morgan:
It seems like this is working and speeding things up dramatically.


resources/config.edn
```diff
@@ -12,7 +12,7 @@
                        :prefetch-limit #resource-config/edn #resource-config/env "SCHEDULED_MESSAGE_WORKS_SCHEDULE_MESSAGE_PREFETCH"
                        :threads #resource-config/edn #resource-config/env "SCHEDULED_MESSAGE_WORKS_SCHEDULE_MESSAGE_THREADS"}
                       {:queue "scheduled-message-works.message.schedule-faf"
```
tank157: I know this queue is for `fire-and-forget`, but considering all the work, I'm going to consider this suffix is for `fast-af`

src/scheduled_message_works/entities/message.clj
```diff
@@ -223,24 +220,42 @@
...
-                                   max-identifiers
-                                   identifier-keys]]))
-           tx-data (->tx-data (assoc message :identifiers-hash identifiers-hash))
-           id (some :scheduled-message/id tx-data)
-           tx-result @(db/transact tx-data)]
-       (entity->message (find-by-id (:db-after tx-result) id))))))
+               (let [identifiers-tx [[:scheduled-message/create-identifiers
+                                      max-identifiers
+                                      identifier-keys]]]
+                  (d/transact-async (db/connection) identifiers-tx)))
```
gws: Since this is a rare call that would only occur as a completely new service was "warming up", should it look like the other schema-level transaction calls and use `d/transact`? I don't see a reason to trade safety here (since this is an async transaction that we don't wait on - `d/transact-async` is not necessarily unsafe).

cap10morgan:
@gws Let me know what you think of that last commit when you have a sec. Thanks!

src/scheduled_message_works/entities/message.clj
```diff
@@ -223,24 +220,43 @@
...
-           tx-data (->tx-data (assoc message :identifiers-hash identifiers-hash))
-           id (some :scheduled-message/id tx-data)
-           tx-result @(db/transact tx-data)]
-       (entity->message (find-by-id (:db-after tx-result) id))))))
-
-(defn-traced update-message [existing-message {:keys [identifiers] :as new-message}]
+               (let [identifiers-tx [[:scheduled-message/create-identifiers
+                                      max-identifiers
+                                      identifier-keys]]]
+                  @(d/transact-async (db/connection) identifiers-tx)))
```
theJohnnyBrown: Is calling deref immediately on the result of `transact-async` the same as using the blocking `transact`?

src/scheduled_message_works/entities/message.clj
```diff
@@ -223,24 +220,43 @@
...
+         (entity->message (find-by-id (:db-after tx-result) id)))))))
+
+(defn schedule-message-async
+  [{:keys [identifiers] :as message}]
+  (s/validate schedule-message-schema message)
+  (let [identifiers-hash (->identifiers-hash identifiers)
+        identifier-keys (map key->identifier-attribute (keys identifiers))
+        _ (when (seq (missing-identifiers (db/db) identifier-keys))
+            (log/info "Adding missing identifiers:" (pr-str identifier-keys))
+            @(d/transact-async (db/connection)
```
theJohnnyBrown: Same as above

src/scheduled_message_works/handlers.clj
```diff
@@ -82,6 +83,24 @@
...
+        hash (hash-content content)]
+    (put! fs/store-file-ch {:location hash :content content})
+    (let [match (message/find-by-identifiers (db/db) (:identifiers message-data))]
+      (if (nil? match)
+        (message/schedule-message-async (-> message-data
+                                            (dissoc :message)
+                                            (assoc :content-hash hash)))
+        (message/update-message-async match (-> message-data
+                                                (dissoc :message)
+                                                (assoc :content-hash hash)))))))
```
theJohnnyBrown: This is the race condition we discussed in slack, where two messages with the same identifiers could be created if they come in close enough to each other. Suggestion was to use a transaction function.

src/scheduled_message_works/entities/message.clj
cap10morgan: Not quite. It eliminates the timeout, which is nice for us b/c we have other timeouts at higher levels of our logic. So as long as those haven't fired, we don't want `transact` to prematurely give up on a transaction. And even when it did that, the transaction would usually go through later anyway. So this seemed to better match the actual behavior too.

src/scheduled_message_works/handlers.clj
cap10morgan: Yeah, inside that `update-message-async` fn (and the non-async one too for that matter) where it queries an arbitrary database value for the existence of a scheduled message w/ those identifiers and then bases its transaction on that. Other transactions may have changed things since that db value was generated. This is a pretty common rookie mistake we made a lot w/ Datomic early on, so I'm not totally shocked to find it lurking in here. I'm going to turn it into a transaction function in a future PR (hopefully today or tomorrow).

src/scheduled_message_works/entities/message.clj
tank157: Is dereferencing the future enough? I mostly ask because it seems like there's a function built for this, `sync-schema`. I'm not quite sure if it's better to use for this case or not.

cap10morgan: It's enough for what we want here (i.e. blocking until the transaction completes or fails). `sync-schema` is different. It's for acquiring database values that contain all schema changes up to a certain time have been transacted.

PR #61, opened 2018-04-03
helper functions to find duplicate scheduled messages

zonotope:


dev-src/scheduled_message_works/duplicate.clj
```diff
@@ -0,0 +1,34 @@
...
+  (:require [scheduled-message-works.entities.message :as message]
+            [datomic.api :as d]))
+
+(defn all-messages [db send-after]
+  (->> (d/q '{:find [[?m ...]]
+              :in [$ ?start]
+              :where [[?m :scheduled-message/send-after ?t]
+                      [(>= ?t ?start)]
+                      [?m :identifier/recipient-id]
+                      [?m :identifier/election-id]
```
cap10morgan: Should we drop `election-id` from this query so we can also find dupes that don't reference an election (e.g. signup messages)?

dev-src/scheduled_message_works/duplicate.clj
```diff
@@ -0,0 +1,34 @@
...
+                      [?m :identifier/notification-type]
+                      [?m :identifier/medium]]}
+            db send-after)
+       (map (partial d/entity db))))
+
+(defn identifier-group [message]
+  (select-keys message [:scheduled-message/send-after
+                        :identifier/recipient-id
+                        :identifier/notification-type
+                        :identifier/medium]))
```
gws: I think you'll also want `:identifier/election-id` here as well and maybe even `:scheduled-message/send-to`, because the query above will pull all messages after `:scheduled-message/send-after`, which could cause users who have scheduled messages for more than one election to be mistakenly grouped together.

I also am not 100% sure that `:scheduled-message/send-after` is a good "grouping" criterion (though I dig putting that part in the query), just because it seems like those could have values that range throughout a given day.

cap10morgan: @gws Not sure if this affects your feedback or not, but we're using matching values of `:send-after` here to detect instances of this bug where the exact same noty messages get scheduled more than once. So we're intentionally trying to ignore even slight variations in those values as legitimate additional scheduled messages for that user. Does that make sense / seem right to you?

gws: The reason for `:scheduled-message/send-to` being a grouping criterion is in case there are `vote-tomorrow` updates in there as well as `upcoming-election` updates - they are separate messages so we'd want them to be counted separately for duplicate identification purposes, rather than counting both with the same user and then deleting an arbitrary message.

gws: @cap10morgan I see, I think I was (and maybe still am) viewing this problem slightly differently. If we follow your first comment above and ignore mine here, I think we'll achieve similar goals :+1:

theJohnnyBrown: Using send-after to group should be fine out of abundance of caution, it's been the same scheduled time for the dupe messages we've seen so far. Leaving out `election-id`, however, is probably not what you want, assuming the plan is to delete all but 1 scheduled noty per group

zonotope: @gws the `vote-tomorrow` and `upcoming-election` messages should show up under the `:identifier/notification-type` key, so those messages should be grouped to different buckets here. 

i left out `:identifier/election-id` because we also saw instances where users got multiple notifications scheduled for about the same time, but the notifications had different election ids, and many of those election ids were for phantom elections that didn't show up in the election-works db. i figured that it was a reasonable assumption that no user would have messages of the same type for legitimately different elections scheduled to be sent on the same day, but that assumption might not be valid.

dev-src/scheduled_message_works/duplicate.clj
zonotope: good call, although dupe signup messages go out so fast that we probably won't catch them, but i'll take out election id.

dev-src/scheduled_message_works/duplicate.clj
cap10morgan: Hmm... I'm not seeing the issue with dropping `election-id`. In general I don't _think_ that the same `notification-type` over the same `medium` for the same `recipient-id` scheduled to send at the exact same time should ever find any dupes, `election-id` or not. If there is more than one legitimate election in the system affecting a given user, it's `:send-after` values should be different because it is (by definition) on a different day. But please poke holes in this logic if there are any to poke!

gws: @zonotope That seems reasonable. I have been thinking of this problem in a specific way, so it took me a bit longer to understand the different approach. Thanks for the discussion everyone! I would just encourage some caution here and some manual review of the data before anything gets thrown in the trash.

zonotope: @gws absolutely! i'm very skittish about prod data cleanups like this. i'll make sure to get as many eyes as possible on the data before issuing transactions.

gws: @zonotope The only other thing I'd consider before trashing any notifications would be to validate that any messages with election IDs slated for deletion are also verified against the election database to ensure that if there is a choice between nuking a notification with an election ID that is a phantom ID vs. one that exists in the database, that we trash the message with the phantom ID. I don't know if these phantom election IDs play into the whole "unknown electorate" thing we see in the logs occasionally, but it seems potentially related to me, a non-TV developer.

zonotope: although this discussion just brought to mind another potential issue with this algorithm and dropping election ids. there could be users who have multiple notifications because of the phantom election id issue. this algorithm is just keeping the first message and will retract the rest. it could be the case that the first message is from the phantom election, and the message for the real election comes later. because of that, we probably want to keep the election id in the query and deal with the phantom election problem separately. i have to think about this more.

zonotope: heh jinx @gws. looks like we were thinking the same thing. i think i'm either going to look for the first dupe with an election id in the election-works db, or i'm just going to handle the phantom election issue separately if the other approach is too slow.

theJohnnyBrown: How far would it get us if we dropped all scheduled messages identified by a "ghost" nonexistent election ID?

zonotope: @theJohnnyBrown i think it would get us pretty far, but we'd have to find all of those. i just pushed a commit which does that at once, but i'm happy to separate the steps if you think that's safer.

cap10morgan:
@zonotope This one is good to merge (and pretty safe), no?

zonotope:
ahh, yes it is. i think i just did the chore without ever merging these. i'll merge them now in case we need code like this again.

PR #62, opened 2018-04-03
k8s changes

cap10morgan:
Merge k8s changes to master. This all already running on the production k8s cluster, and this branch should be ready to run on Fleet clusters too now. I've added the new values to the tvprod Consul store.


src/scheduled_message_works/entities/message.clj
```diff
@@ -137,55 +145,68 @@
...
+             (all-identifier-keys? db identifiers))
+    (let [;; build up a where clause which matches messages (named
+          ;; `?e`) that matches all keys and values from identifiers
+          all-identifiers-match
+          (build-identifiers-query identifiers)
+
+          query {:find '[?e]
+                 :where all-identifiers-match}
+          entities (d/q query db)]
+      (->> entities
```
ddellacosta: @cap10morgan Per our slack discussion think you could change this to use a transducer and get a bit more efficiency, not that it matters much here probably:

```clojure
(eduction
  (comp
    (map first)
    (map (partial d/entity db))
    (map entity->message))
  entities)
```

Of course, it's probably simpler to just compose the internal functions with a single `map`!

```clojure
(map (comp entity->message (partial d/entity db) first) entities)
```

ddellacosta: Oh whoops, sorry I missed that...in any case at least it provided a topic of conversation!

cap10morgan:
@ddellacosta How does that look (last commit)?

ddellacosta:
@cap10morgan I didn't think to tweak the query to return the single result--even better!

src/scheduled_message_works/entities/message.clj
cap10morgan: I learned some things about `eduction` while playing around with this::

1. It takes one or more transducers, so you don't even need the `comp`.
1. It returns an `Eduction` instance that remembers how to apply the transducer(s) to the collection you give it. So you have to do something with it to turn into the data structure you want.

So... I ended up going with your simpler suggestion in the end.

Thanks for the learnings, though!

ddellacosta: Cool, I did not know either of those facts!

ddellacosta: @cap10morgan Actually I'm confused by the second point maybe because I'm running this in the REPL: are you saying you can re-use the `Eduction` return value in some way?

```clojure
user> (def eduction-test (eduction (comp (map #(+ 1 %)) (map #(* 2 %))) [1 2 3]))
#'user/eduction-test
user> eduction-test
(4 6 8)
user> (type eduction-test)
clojure.core.Eduction
user>
```

PR #64, opened 2018-05-14
add queue for user ready events

theJohnnyBrown:
Needed for [user-works to request events](https://www.pivotaltracker.com/story/show/156697443) and export to ActionKit


src/scheduled_message_works/channels.clj
```diff
@@ -30,7 +31,10 @@
     ballot-delivery-update
 
     "election-notification-works.signup.send"
-    signup-send} queue))
+    signup-send
+
+    "user-works.user.ready"
+    user-ready} queue))
 
 (defn close-all! []
```
gws: The `user-ready` channel should be closed in this function as well.

mrichards42:
I don't have a good answer for where this belongs, but I'm not sure it's s-m-w. Do you have thoughts on pros or cons of putting it here?

mrichards42:
I guess the implementation of this event is going to be based on a delay, so that's a decent reason to put it here, rather than requiring an extra trip back to another service just so it can fire this event.

PR #65, opened 2018-05-17
use events exchange for events

theJohnnyBrown:


mrichards42:
Looking at other services that send messages to the events exchange, I think you also need to declare the event exchange in the top-level kehaar config like so: https://github.com/democracyworks/user-works/blob/master/resources/config.edn#L4

Also, looking at that same resources/config, it looks like `events` routing-keys aren't ns-qualified (I assume since they aren't being sent to a particular queue), so the routing-key should be `user.ready`. I think the other s-m-w outgoing-events are special cases since they're being delivered to a particular queue instead of the events exchange.

theJohnnyBrown:
@mrichards42, I hear you about the namespace qualifier. I'll change that here and in user-works.

As far as the exchange, I've verified that it works with the exchange specified this way.

PR #66, opened 2018-08-30
[WIP] rok8s-scripts, build, deploy refactor

democracyworks-reactiveops:

PR #69, opened 2018-09-17
Reconcile deploys

cap10morgan:
This is an attempt to land `ro-deploy` on `master` and leave it able to deploy to the Ballot Scout fleet cluster as well.


.buildkite/pipeline.yml
```diff
@@ -1,31 +1,26 @@
...
 
-  - trigger: democracy-works-kubernetes
-    label: Deploy
-    branches: master
-    async: true
-    build:
-      branch: master
-      env:
-        DW_COMPONENT: scheduled-message-works
-        SOURCE_JOB_ID: "${BUILDKITE_JOB_ID}"
```
gws: If we were to leave this in but remove the old TurboVote Kubernetes cluster from the `dw-k8s` repo, we'd get Election Reminders (Gov Noties) deploying as well, I think? The only hidden dependency that triggered pipeline has is the image name that got built gets set in the Buildkite metadata, so as long as the existing `script/build` can do that, that should work too...

gws: To make it more concrete, if something like this can happen where that meta-data key gets set with the full image name + tag during the build step, everything should work:

https://github.com/democracyworks/scheduled-message-works/blob/master/script/build#L72-L73

cap10morgan:
@gws This look about right?

script/build
```diff
@@ -1,74 +1,26 @@
 #!/bin/bash
 
-set -e
```
jlwynkoop: I think we still need to `set -o errexit` so that rok8s-scripts will cause the buildkite job to fail if a deploy fails.

deploy/scheduled-message-works-service-monitor.service.yml
```diff
@@ -0,0 +1,20 @@
+# We're including a servicemonitor here because rok8s-scripts
```
jlwynkoop: It's worth leaving this for now, but we'll probably be able to rip it out once external metrics are working.

docker/app_container/usr/local/bin/entrypoint.sh
```diff
@@ -0,0 +1,22 @@
...
+        -XX:+UnlockExperimentalVMOptions \
+        -XX:+UseCGroupMemoryLimitForHeap \
+        -XX:MaxRAMFraction=${MaxRAMFraction:-2}"
+    else
+    echo "JVM Heap statically set"
+      export JVM_LINE="-Xms${XMS:-1g} -Xmx${XMX:-1g}"
+fi
+
+
+java ${JVM_OPTS:--XX:+UseG1GC} \
```
jlwynkoop: I think we're missing an argument here to exit on OOO errors.

script/build
democracyworks-reactiveops: for some reason this is being cached by github. but in git ... it's fine

democracyworks-reactiveops:
closing in favor of #66

PR #70, opened 2018-09-21
Replace transact w/ transact-async everywhere

cap10morgan:
This might cut down on some of the error noise we're getting when load is high. Often these transactions succeed anyway, and the timeout from `transact` is an arbitrary value we haven't spent any time tuning.


mrichards42:
Closed since this was done in #74

cap10morgan:
Oh that's funny. Thanks for the clean up!

PR #71, opened 2018-09-23
Duplicate notys bug

li-at-dw:
https://www.pivotaltracker.com/n/projects/930636/stories/159277120

It was discovered when reviewing the logs that it is possible (due to various factors) for a node to  work on sending out messages for a length of time which exceeds the reservation time limit ( "SCHEDULED_MESSAGE_WORKS_UNRESERVE_OLDER_THAN_IN_MINUTES").  Once the reservation time limit has passed, other nodes will invalidate the reservation, assuming that "something bad" has happened to the node which made the original reservation.  This causes a situation where a message can be sent multiple times by different nodes cancelling reservations made by other nodes.  

This ticket is intended to fix this, by making sure that the process for sending scheduled messages quits when the reservation time limit has passed.


test/scheduled_message_works/send_messages_test.clj
```diff
@@ -114,6 +124,49 @@
     (is (= 2 (count @messages)))
     (is (= #{{:message 1} {:message 2}} @messages))))
 
+
+(deftest timeout-test
+  (comment
```
cap10morgan: Should we remove this now?

li-at-dw: yes!  Thanks!

src/scheduled_message_works/send_messages.clj
```diff
@@ -36,13 +36,19 @@
...
-  [channel-fn sendable-messages]
+  unhydrates scheduled-messages that are sendable, sends them
+  if current time is before the deadline."
+  [channel-fn sendable-messages deadline]
   (when (seq sendable-messages)
     (dorun (map-indexed
             (fn [idx message]
-              (log/debug "sending message number" (+ 1 idx))
-              (send-message channel-fn message))
+              (if (time/before? (time/now) deadline)
```
mrichards42: I wonder if this leaves us open to another small race condition: since `send-message` (and the `delete-message` afterwards) aren't instantaneous, another scheduler could reclaim this message when `deadline` is very close to now, but still after now.

There probably isn't a perfect solution, but what about having this stop sending when the current time is within a certain threshold of the deadline (say, a few seconds)?

li-at-dw: That's a good point.  I looked through the logs a bit more, and the longest time for sending a message was just above 1 minute.  So, I would propose to cut the deadline short by maybe 2 minutes, just be safe....

test/scheduled_message_works/send_messages_test.clj
```diff
@@ -22,6 +22,16 @@
...
 
+(defn take-until-timeout
+  ([chan]
+   (take-until-timeout chan (atom #{})))
+  ([chan result]
+   (let [curr (take-or-timeout chan)]
+     (swap! result conj curr)
+     (if (= :timeout curr)
+       result
+       (recur chan result)))))
```
mrichards42: It took me a little while to understand the tests needed to check against `(+ 1 num-messages)` and I realized ultimately it was b/c `:timeout` is included in the return from this function. If you wanted to exclude `:timeout` you could do the `swap!` inside recur, like this:
```clj

(defn take-until-timeout
  ([chan]
   (take-until-timeout chan (atom #{})))
  ([chan result]
   (let [curr (take-or-timeout chan)]
     (if (= :timeout curr)
       result
       (recur chan (swap! conj result curr)))))) 
```
though actually, at that point you don't need `result` to be an atom at all:
```clj
;; assuming result is a set instead of an atom
(recur chan (conj result curr))
```

test/scheduled_message_works/send_messages_test.clj
```diff
@@ -114,6 +124,48 @@
...
+(deftest timeout-test
+  (testing "time out immediately"
+    (let [outbound (async/chan)
+          num-messages 1000
+          messages-in (map #(td/test-message-entity (date-offset -20000)
+                                                    {:content-hash
+                                                     (->content-hash {:message %})})
+                           (range num-messages))
+          scheduled (doseq [x messages-in]
+                      (message/schedule-message x))
```
mrichards42: If you're only calling a single function in`doseq` you can use:
```clj
;; Like doseq, this returns nil
(run! message/schedule-message messages-in)
```
or if you need the results (and you don't want it to be lazy):
```clj
(mapv message/schedule-message messages-in)
```

test/scheduled_message_works/send_messages_test.clj
```diff
@@ -114,6 +124,48 @@
...
+          messages-in (map #(td/test-message-entity (date-offset -20000)
+                                                    {:content-hash
+                                                     (->content-hash {:message %})})
+                           (range num-messages))
+          scheduled (doseq [x messages-in]
+                      (message/schedule-message x))]
+      (reclaim-and-send num-messages 1 (constantly outbound) (time/now)
+                        (time/plus (time/now) (time/seconds 1)))
+      (let [messages (take-until-timeout outbound)]
+        (is (< (count @messages) num-messages))))))
```
mrichards42: TL;DR -- time is hard in tests

I wasn't trying to write an essay but it sort of ended up that way. Like I said in the overall review comment, this isn't a show-stopper, just some ideas that came out of thinking about how to test this code.

Tests that rely on time are a fragile by nature (e.g. if we run the tests on a really fast machine that can get through all 10k messages in a second, this test fails). Since we're using `clj-time` you can use the `do-at` macro to remove time as a variable:
```clj
(time/do-at
  (time/date-time 2016 6 7)
  ;; Everything inside this block will use a fixed time of 2016-06-07T00:00:00
  ...
  (org.joda.time.DateTimeUtils/setCurrentMillisFixed
    (.getMillis (time/date-time 2016 6 7 1)))
  ;; Now time is fixed at 2016-06-07T01:00:00
  ;; This isn't documented in clj-time, but it's what `do-at` uses internally
  ...
  )
;; After this form, time is reset to the actual system time (and it keeps moving)
```

Unfortunately there isn't a clean way to send half the messages, then change the time, then try to send the other half of the messages. 

Also `clj-time` is effectively deprecated with java 8's `java.time` (we'll likely migrate clj-time code eventually, but it isn't a high priority), and `java.time` doesn't make it easy to set time to a fixed point for your tests. Instead you would have to pass in a function for getting the current time, which is at least a clean solution, but still pretty annoying:
```clj
(defn send-scheduled-messages
  [channel-fn sendable deadline now-fn]
  ...
  (if (time/before? (now-fn) deadline)
    ...))

(defn reclaim-and-send
  [batch-size minutes-ago channel-fn time deadline now-fn]
  ...
  (send-scheduled-messages channel-fn sendable deadline now-fn))

;; Now you can control time in your tests
(defn inc-time-fn [now increment]
  (let [now-atom (atom now)]
    (fn [] (swap! now-atom time/plus increment))))

(let [now (time/now)]
  (reclaim-and-send 1 (constantly outbound)
                    now (time/plus now (time/seconds 100))
                    (inc-time-fn now (time/seconds 1))))
```

PR #75, opened 2018-10-04
Further speed up reserved message using raw index access

gws:
Rather than potentially fetch all messages in the database, we can lazily walk the `:scheduled-message/send-after` index for all messages before the time provided, and just take the first `n` messages that have not already been reserved.


resources/schemas/2018-10-04-001-speed-up-reserve-messages.edn
```diff
@@ -0,0 +1,29 @@
...
+                (datomic.api/entity db e)))
+         without-reservation-xf
+         (filter (fn [entity]
+                   (nil? (:scheduled-message/reservation-id entity))))
+         as-tx-data-xf
+         (map (fn [entity]
+                {:db/id (:db/id entity)
+                 :scheduled-message/reservation-id reservation-id}))
+
+         batch (take at-most (sequence (comp as-entity-xf
```
tank157: Is there some reason not to make the `(take at-most)` also a transducer and put it in before the `as-tx-data-xf` so we only map as many as we're actually going to use?

gws: It's not necessary - `datoms` is lazy (well, it's an Iterable), so we are only walking as far through the filtered sequence as necessary to satisfy `(take at-most ...)`.

gws: Unless I'm missing something else about what you're saying? I want to be careful.

tank157: Oh, I suppose that's probably correct, I'm not so used to working with transducers still, but yeah, it it's essentially lazy all the way down to the `take` then it's probably fine.

PR #77, opened 2018-11-01
Reserve stagnant messages

mrichards42:
This is a significant change to how we reserve messages, so if it seems like too much right now, I'm happy to defer this and open a separate PR that takes on the smaller task of limiting the number of messages that can be reserved at one time.

The main idea: rather than unreserve stagnant messages and _then_ reserve messages, make stagnant messages available to be reserved.

The other idea: `:reservation-id` already has a timestamp embedded in it, since it's a squuid. This code takes advantage of the timestamp to avoid the extra hops to `tx` and then `txInstant`.


gws:
What about modeling this directly with an indexed `reserved-at` attribute? It may seem unlikely to be a problem right now but I think directly modeling this concept will end up being less painful in the future.

mrichards42:
I can see using `reservation-id` _as_ the timestamp making this a little fragile (in fact, I had to jump through a couple hoops while writing tests), so I can definitely see using a real timestamp as a better plan. Does `reserved-at` give us anything more than `txInstant` does?

gws:
In the sense that they're both timestamps you have access to, no they're not different. However, in my experience, using `txInstant` for day-to-day work indicates something missing in your data model. It's tempting to use, but I think it's best kept to one-off [audit trail queries using the history filter](https://docs.datomic.com/on-prem/best-practices.html#use-history) and that sort of thing. I can't really elaborate more right now but it's caused pain and suffering in the Ballot Scout codebase where it is relied upon extensively, so that's mainly the experience I'm speaking from (and also from anecdotes on Datomic mailing lists, etc.)

mrichards42:
I went ahead and put up #78 as the simpler alternative.

@gws now that I'm thinking about it, why not use `reserved-at` instead of `reservation-id`? I don't think we actually need the id, aside from as an indicator that something is reserved. We were using it in `entities.reserve-messages` to query for messages that were reserved in the batch, but since the transaction returns all the datoms, you can get that info elsewhere. Not that there's a problem with having both `reserved-at` and `reservation-id`, but it does seem like there's some redundancy.

PR #78, opened 2018-11-01
Unreserve batch

mrichards42:
Simpler than #77 but still achieves the goal of limiting transaction sizes.


resources/schemas/2018-11-01-001-unreserve-batch.edn
```diff
@@ -0,0 +1,25 @@
...
+
+         as-tx-data-xf
+         (map (fn [{:keys [e v]}]
+                [:db/retract e :scheduled-message/reservation-id v]))
+
+         batch (take at-most (sequence (comp with-reservation-before-xf
+                                             as-tx-data-xf)
+                                       datoms))]
+     (when (seq batch)
+       batch))}}]
```
gws: This is one of things that I only realized *after* I touched the code (and I know you didn't put it there!), but

```clojure
(when (seq batch)
  batch)
```

can just be

```
(seq batch)
```

... they both return `nil` when `batch` is empty or the sequence otherwise, I'm pretty sure they're equivalent for our purposes.

mrichards42: I don't think it matters if this returns an empty sequence or nil actually, so just `batch` should work. Both `nil` and `[]` seem to record just the transaction data:

```clj
=> (:tx-data @(d/transact-async dummy-connection nil))
[#datom[13194139534333 50 #inst "2018-11-02T14:52:27.631-00:00" 13194139534333 true]]
=> (:tx-data @(d/transact-async dummy-connection []))
[#datom[13194139534334 50 #inst "2018-11-02T14:52:31.654-00:00" 13194139534334 true]]
```

gws: Maybe it wasn't this function, but there are some places that *just* check the return value without nil-punning (maybe just for `reserve`), and empty collection by themselves aren't falsey. Even better if this works, though!

PR #80, opened 2019-03-06
Continuous sending

mrichards42:
This is a rewrite of `scheduler/start` that allows s-m-w to send continuously as long as there are messages available.

Based on logs, we're able to send the current (turbovote) batch-size of 250 in about 10-30 seconds. At a granularity of 2 minutes that's, at minimum, 75% idle time. This would also allow us to set granularity higher, since granularity is only enforced when there _aren't_ scheduled messages available.

Putting a bunch of people on this PR since this cuts across most of our teams.


mrichards42:
The `scheduler/start` tests are a little wonky since they depend on the current time in several places. Definitely open to feedback on better ways to write those tests.

src/scheduled_message_works/scheduler.clj
```diff
@@ -1,32 +1,37 @@
...
-(defn periodic-sequence
-  "Given a start time and a periodic granularity,
-  returns a periodic-sequence of times at the granulatity
-  intervals from the start time."
-  [start granularity]
-  (periodic/periodic-seq start (time/minutes granularity)))
+(defn ms-between [start end]
+  (if (time/before? end start)
+    0
+    (-> (time/interval start end) (time/in-millis))))
```
tank157: I was thinking you could maybe also get away with the `(max 0 (time/interval start end))` trick here, but it looks like Interval will throw an IllegalArgumentException if the end is before the start.

src/scheduled_message_works/scheduler.clj
```diff
@@ -1,32 +1,37 @@
...
     (async/go-loop []
-      (when-let [time (async/<! chimes)]
-        (process-fn time)
+      (when-let [time (async/<! time-ch)]
+        (let [wants-more? (= :wants-more (process-fn time))
+              timeout-ms (if wants-more?
+                           0
+                           (next-timeout granularity-ms time (time/now)))]
+          (when (pos? timeout-ms)
+            (async/<! (async/timeout timeout-ms)))
```
tank157: So it's really this timeout that's doing the "wait for the next time" sort of thing? It's all a bit confusing to read, but I don't know of another way to handle it offhand.

src/scheduled_message_works/scheduler.clj
mrichards42: I actually grabbed this function straight from [chime](https://github.com/jarohen/chime/blob/f28d658bc5fc14f49c929d28718929194760bf20/src/chime.clj#L7-L10)

src/scheduled_message_works/scheduler.clj
mrichards42: right, this line is essentially "park until the next time"

Now that I think about it, I actually think the majority of this PR could have been:
```diff
  (defn start
    [granularity process-fn]
    (let [time-seq (periodic-sequence (time/now) granularity)
          buffer (async/chan (async/sliding-buffer 1))
          chimes (chime/chime-ch time-seq {:ch buffer})]
      (async/go-loop []
        (when-let [time (async/<! chimes)]
-         (process-fn time)
+         (when (= :wants-more (process-fn time))
+           (async/>! chimes (time/now))
          (recur)))
      chimes))
```

I think the added complexity in this PR amounts to implementing a tiny version of `chime`

gws:
@mrichards42 If you're using `clj-time` exclusively, you can use [`clj-time.core/do-at`](https://github.com/clj-time/clj-time/blob/master/src/clj_time/core.clj#L774-L778):

```clojure
(deftest your-test
  (testing "becoming a :timelord:"
    (time/do-at (time/date-time 2018 05 02 14 1)
      (is (= :sent (some-expr-that-uses-clj-time))))
```

When I'm using the `java.time` library directly, I typically bind a `Clock` to a dynamic var like `*clock*`, which is simply the regular old system clock by default, but you can re-bind that in tests to fix the clock to a specific point in time. In the actual namespace, all your calls to `now` talk to `*clock*` to figure out what that is.

mrichards42:
A couple of changes:

1. f4e01a9 allows stopping the scheduler at any point. Without the `alts!` call the scheduler would remain parked until the timeout elapsed, which probably isn't a huge deal, but this is still better.

2. @tank157 @gws I reworked tests to inject time fns. You both hit on one of the difficulties I was having (injecting `time/now`), but the bigger problem was getting `async/timeout` consistent across tests. I think the result is a lot better, but it does involve a lot of mocking in the tests.

gws:
Edited to provide an example of a possibly-simplified but untested scheduler.

mrichards42:
Based on gordon's comments above and a fruitful discussion in slack (starting [here](https://democracyworks.slack.com/archives/C0FUNUSGN/p1552486233002200)), the latest commit is an even simpler version of the scheduler.

mrichards42:
@li-at-dw I actually think this will help with the problem, since this should mean that we won't have to run as many replicas

mrichards42:
The approach I ended up with is different enough that I'm going to close this and re-open as a new PR for a fresh review

PR #82, opened 2019-04-01
add queue/channel for ballot reminders

theJohnnyBrown:
Need for this story: https://www.pivotaltracker.com/story/show/164616079

Also, thoughts on using langohr instead of kehaar when sending the scheduled messages? It would remove the need for a code change to pre-declare every new routing key we want to use, clients could specify the routing key and langohr could send the message to whatever key was given.


theJohnnyBrown:
@katrinamariehh thanks for the pointer, I thought it might have come up before. Could be a simplifying change whenever someone has the bandwidth for it.

katrinamariehh:
👍

PR #86, opened 2019-09-05
Rewrite scheduled-message-works to use PostgreSQL

gws:
This is a breaking change for existing deployments, as the service now requires a PostgreSQL database as opposed to a Datomic database.

The changes are large because a lot of unnecessary code has been removed, so it's probably easiest to poke around at the code in the branch. I hope it is as readable as I think it is!

Some notes:

- Sending messages can now occur as fast as RabbitMQ can accept them, and polling uses indexed queries and so is very cheap. We should tune our sending based on message volume rather than out of scalability concerns.
- API-wise, Response attributes such as content and identifier hashes are no longer returned in the response (since they no longer exist), but none of our systems care about those attributes (in fact, most of them don't even look at the s-m-w responses anymore).


src/scheduled_message_works/db.clj
```diff
@@ -0,0 +1,214 @@
...
+          (sql/send-message tx (:id msg) (Instant/now))
+          (catch Exception e
+            (log/errorf "Failed to send message ID %s: %s"
+                        (:id msg)
+                        (.getMessage e))
+            (metrics/inc-failed-messages! registry)
+            ;; Will throw and roll back the transaction if it fails, potentially
+            ;; re-sending messages, but if it fails due to a runtime error (the
+            ;; most likely scenario) the transaction almost certainly wouldn't
+            ;; be able to commit and would be rolled back anyway.
```
mrichards42: I can't say I fully understand how likely this would be, but given how much work has gone into reducing duplicate scheduled messages, the idea that certain errors could cause an entire batch of messages to be re-sent makes me nervous. In particular, my expectation is that any time a service is terminated (i.e. any time we redeploy) this transaction be rolled back and any messages that were already sent in this batch would be re-sent.

You've been telling me that I don't need to worry so much about message timing and the like given how much faster this system is, so it might be that given an appropriate `limit` this isn't a concern in reality. I know the older system -- and honestly many of our systems -- work essentially one row at a time (thinking about how each scheduled message was deleted immediately after it was sent in the old system). In theory that would nearly eliminate this class of problems, but is probably horribly inefficient.

src/scheduled_message_works/db/sql.clj
```diff
@@ -0,0 +1,135 @@
...
+   ["SELECT m.id, m.send_after, m.send_before, m.content, m.identifiers,
+            ma.routing_key
+     FROM message AS m
+     JOIN message_amqp AS ma
+     ON m.id = ma.message_id
+     WHERE m.identifiers @> ?::jsonb AND m.state = 'scheduled'"
+    identifiers]))
+
+(defn scheduled-message-count
+  "Returns a count of scheduled messages."
```
mrichards42: Are we using this function anywhere? I could see something like this being useful in reporting, but usually we care about counts per day or hour or something like that, rather than a total count.

src/scheduled_message_works/db/sql.clj
```diff
@@ -0,0 +1,135 @@
...
+  "Delete up to one message that exactly matches `identifiers`."
+  [connectable identifiers]
+  (jdbc/execute-one! connectable
+   ["DELETE FROM message
+     WHERE id = (
+       SELECT id FROM message
+       WHERE identifiers = ?::jsonb
+         AND state = 'scheduled'
+       LIMIT 1
+       FOR UPDATE
```
mrichards42: I puzzled over this subquery for a while, but at this point I assume it's there b/c `DELETE ... LIMIT 1` isn't a thing? Do we need the limit here, since as I understand the model, identifiers should be unique?

src/scheduled_message_works/metrics.clj
```diff
@@ -0,0 +1,34 @@
+(ns scheduled-message-works.metrics
```
mrichards42: 👏

src/scheduled_message_works/metrics.clj
```diff
@@ -0,0 +1,34 @@
...
+            [iapetos.standalone :as web]
+            [mount.core :refer [defstate]]))
+
+(def failed-message-counter
+  (prometheus/counter :message/failed
+   {:description "Number of messages we failed to send."}))
+
+(def sent-message-counter
+  (prometheus/counter :message/sent
+   {:description "Number of messages we sent successfully."}))
```
mrichards42: Would it be helpful to have these counts broken down by `routing_key`? We have a relatively limited number of queues we send to.

src/scheduled_message_works/queue.clj
```diff
@@ -1,41 +1,33 @@
...
-  (let [max-retries 5
-        rabbit-cfg (dissoc-blank-config-params-with-defaults connection-config)
-        connection (kehaar.rabbitmq/connect-with-retries rabbit-cfg max-retries)
-        kehaar-resources (kehaar/init! connection kehaar-config)]
-    {:connections [connection]
+  (let [{kehaar-cfg :kehaar connection-cfg :connection} config
+        rabbit-cfg (k.rabbitmq/dissoc-blank-config-params-with-defaults
+                    connection-cfg)
+        connection (k.rabbitmq/connect-with-retries rabbit-cfg max-retries)
+        kehaar-resources (k.configured/init! connection kehaar-cfg)]
```
mrichards42: I forget, but I seem to remember having ordering issues using mount w/ kehaar.configured _and_ stateful message handlers, since mount has no way of knowing that kehaar needs to be started after the handlers (since those connections are defined in a config file). Did this not come up for you?

doc/postgres-migration.md
```diff
@@ -0,0 +1,34 @@
...
+## 3. ~Draw the rest of the owl~ Perform the migration
+
+At this point, the timetable is short. Messages will be delayed until after this
+piece is done.
+
+1. Scale down s-m-w sending replicas to 0.
+2. Scale down s-m-w no-sending replicas to 1.
+3. Deploy the new s-m-w replicas.
+4. Once one of the new replicas comes up, shut down the final s-m-w no-sending
+   replica.
```
mrichards42: Would it work to tear down all the existing s-m-w and no-sending replicas and let the `scheduled-message-works.message.schedule[-faf]` queues back up while we wait for the new replicas to come up? I guess if there are other services _not_ using `faf` they might start timing out.

Dockerfile
```diff
@@ -12,19 +12,19 @@ WORKDIR /home/scheduled-message-works/src
...
-ARG DATOMIC_PASSWORD
+COPY --chown=scheduled-message-works:scheduled-message-works project.clj ./
 
 RUN lein deps
 
 ## Add the rest of the tree with files that change more often.
 COPY --chown=scheduled-message-works:scheduled-message-works . .
 
-RUN lein do test, uberjar
+RUN lein uberjar
```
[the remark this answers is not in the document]
gws: Yeah, exactly.

mrichards42:
RE: unique identifiers, afaik all turbovote services treat identifiers as if they were unique (e.g. reschedule is find-message-by-identifiers -> delete -> schedule), but I see that it was never enforced in the datomic schema, so 🤷‍♂

resources/sql/001-initial.up.sql
```diff
@@ -0,0 +1,75 @@
...
+ */
+
+-- message
+
+CREATE TABLE IF NOT EXISTS "message" (
+  "id" uuid PRIMARY KEY,
+  "send_after" timestamptz NOT NULL,
+  "send_before" timestamptz,
+  "content" text NOT NULL,
+  "identifiers" jsonb,
```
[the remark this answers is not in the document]
gws: This is a good point, the schema looseness is because I built this to be able to handle the existing data too though, since we're not adding it fresh. It's possible we could tighten this up with some more testing, but the current code does an upsert by deleting *just one* of the matching identifiers and re-inserting, so there's no way to know if the current data would fit within that constraint.

src/scheduled_message_works/db.clj
```diff
@@ -0,0 +1,214 @@
...
+           :message/content :message
+           :message/identifiers :identifiers
+           :message_amqp/routing_key :send-to}]
+    (as-> row $
+      (select-keys $ (keys m))
+      ;; TODO: Move type conversion to the edge; use Instant internally
+      (update $ :message/send_after #(Date/from %))
+      (update $ :message/send_before #(some-> % (Date/from)))
+      (update $ :message/content edn/read-string)
+      (update $ :message/identifiers #(some-> % (db->identifiers)))
```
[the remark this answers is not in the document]
gws: Ah yeah, I think this predates the restrictive spec.

src/scheduled_message_works/db/sql.clj
gws: They might be, but the current code behaves this way, in that it only tries to fetch one set of matching identifiers and deletes those. And yeah, this is how you need to do it to limit the number of deleted rows.

src/scheduled_message_works/db.clj
```diff
@@ -0,0 +1,214 @@
...
+
+  The choice to match a single message was made to match the existing
+  semantics."
+  [ds m]
+  (let [m' (scheduled-message->db m)]
+    (jdbc/with-transaction [tx ds]
+      (sql/delete-message-by-strict-identifiers tx (:message/identifiers m'))
+      (let [ret (merge (sql/insert-message! tx m')
+                       (sql/insert-message-amqp! tx m'))]
+        (db->scheduled-message ret)))))
```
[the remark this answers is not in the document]
gws: This is mimicking exactly the way the current code works (for better or worse) - I wasn't sure if anything cared about message IDs changing (or, here, not changing).

src/scheduled_message_works/queue.clj
gws: I fixed it by ensuring namespaces are required in the right order, `mount` only knows as much information as you give it and you sometimes have to be more cognizant of `require` ordering... I won't say it was obvious to me at first, though.

doc/postgres-migration.md
gws: Yeah, if you leave 1 no-sending it'll just keep handling requests, and the cutover should be seamless. We can do this manually through `kubectl scale`, so it shouldn't be too bad to follow those instructions, but in a pinch you're probably right and that would work as well.

src/scheduled_message_works/queue.clj
mrichards42: Ah, in this namespace in fact. It might be helpful to have a comment above the channels and handlers requires explaining that their states need to be defined before the state in this ns since that part isn't entirely obvious.

src/scheduled_message_works/metrics.clj
gws: Yeah, I think that might be nice to have, would it be a problem to defer this to the future once I get this working with Datadog?

gws: Another thing I was thinking of here was your bar chart - it'd be nice to have graphs like that in Datadog automatically, the query to get that info is pretty straightforward...

resources/sql/001-initial.up.sql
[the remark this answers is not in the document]
gws: I mean, there's a way to know, but I was thinking we could tighten this later, and I just wrote the queries to behave similarly.

src/scheduled_message_works/metrics.clj
mrichards42: Yeah, absolutely. Having these sending and failure rates as graphs will already be amazing. I might missing something obvious, but I haven't thought of a straightforward way to get the _future_ sending graph in datadog, even though the sql for that would be simple.

resources/sql/001-initial.up.sql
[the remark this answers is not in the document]
mrichards42: 👍 totally something we can deal w/ later, and especially given how big a change this is, we'll probably be thankful that the logic is as similar as possible.

src/scheduled_message_works/db.clj
gws: No, actually, the more I think about this the more I think it might even be over-optimized -it really is that fast. I think performing a single transaction at a time is absolutely doable, wouldn't be measurably slower, and I think you're right to prioritize safety here.

I'll give an implementation a shot, it shouldn't take too long.

gws: >In particular, my expectation is that any time a service is terminated (i.e. any time we redeploy) this transaction be rolled back and any messages that were already sent in this batch would be re-sent.

In the case of a controlled shutdown, the current batch should finish sending before the service shuts down if I wrote this correctly.

src/scheduled_message_works/db/sql.clj
gws: This might have been for early REPL-driven development, will double-check and remove if necessary :+1:

src/scheduled_message_works/db.clj
mrichards42: > In the case of a controlled shutdown, the current batch should finish sending before the service shuts down if I wrote this correctly.

I think k8s sends SIGTERM, which I would hope triggers the shutdown hook, so I agree with you there. I'm probably thinking about the fact that the old implementation could take several minutes to get through a batch, and if I understand correctly, k8s sends SIGKILL after 30 seconds, which _was_ a big problem (though as you've said -- sounds like it won't be an issue with this implementation).

src/scheduled_message_works/metrics.clj
gws: I think you could do it with a Prometheus gauge, a labeled metric, and some buckets (such as each day for the next 7, 14, whatever days). You'd run the query in a loop every 5, 10, 30 minutes or whatever to update the gauge. I haven't found a way to do dynamic buckets (e.g. show me how many future scheduled messages are going out, grouped by the day they're going out on), which IMO would be the nicest, but fixed bucketing would be neat.

When/if this gets to production and we're confident with it, I think adding this kind of thing would be straightforward (I think I was using that count function as a toy example to play with gauges). The other neat thing about prometheus is you can test your stuff locally by hitting localhost:9090/metrics since it's a text-based format. How Datadog interprets it is another matter, though.

src/scheduled_message_works/queue.clj
gws: Got it, will be in the next push.

gws:
@mrichards42 I went ahead and implemented continuous sending and did away with batching, since we're doing the safest thing possible now.

- Grab and mark a message for sending
- Attempt to send the message
- Either mark it sent or failed

Otherwise I tightened some things up, and made the concurrency integration tests a little more aggressive, and fixed the other little things.

gws:
It looks like it takes about 9-11ms to send each message now, which should be fast enough.

gws:
I think I'm also going to enforce uniqueness on `identifiers`, I just ran a query on the prod dataset and there aren't any duplicates.

src/scheduled_message_works/db.clj
```diff
@@ -0,0 +1,212 @@
...
+
+  Returns `true` if a message was sent successfully, `false` if it failed, and
+  `nil` to indicate there was no sendable message in the database."
+  [ds f now]
+  (when-let [message
+             (jdbc/with-transaction [tx ds]
+               (when-let [message (some-> (sql/select-message-for-sending tx now)
+                               (db->scheduled-message))]
+                 (sql/grab-message tx (:id message))
+                 message))]
```
mrichards42: I think what you had before was actually a little more readable: wrapping the whole function in a transaction, and having row locks instead of message reservations felt very clean. I'm not opposed to the extra `sending` state, although I think it means we need to find a way to clean up unsent `sending` messages, the same way we had to deal w/ stagnant reserved messages in the old system.

gws: I liked it better too, it just depends on what our risk tolerance is like. I think the risk of duplicate sends is pretty low, because it requires that the DB is available to read a message, then we successfully send it, but then it's unavailable to mark the message sent.

mrichards42: Oh, I think I see what you're getting at. I feel a lot more OK with the risk of unexpected/unexplained errors causing 1 message to be re-sent than it causing an entire batch to be sent, so either adapting the old code to run with a limit of 1, or keeping this as-is and going with a `sending` state is OK by me.

I suppose with this code if we have unexpected errors we'll see messages stuck in `sending`, which is perhaps better than messages being re-sent, but either way I agree that the risk of either happening is quite low, so I'm fine with whichever route you think is best.

gws: I thought about it a little bit and I think we can mitigate the risk of violating some kind of future constraint through marking a message `sent`, causing future send attempts to fail repeatedly (weird, I know, but I'm about to implement the partial unique index on (`identifiers`, `state`) you were hinting at), and to some degree database unavailability, by putting that first in the transaction before performing the side effect.

src/scheduled_message_works/sending.clj
```diff
@@ -0,0 +1,76 @@
...
+          result (async/alt!!
+                   [[ch (:message message)]] true
+                   timeout-ch :timeout)]
+      (when (= :timeout result)
+        (throw (ex-info "Timed out sending message"
+                        {:message message}))))
+    (throw (ex-info "No channels configured to send message"
+                    {:message message}))))
+
+(defn start-sender
```
eggsyntax: I'm pretty unclear on the design here with this, the `send-message` fn it takes, and the `sql/send-message` in `run-with-sendable-message`. Digging down, I think that maybe `sql/send-message` is confusingly named; it seems like it's just setting the state of the message as sent. But maybe also some changes in the docstrings to make those patterns clearer?

src/scheduled_message_works/db.clj
```diff
@@ -0,0 +1,212 @@
...
+  message will be marked as failed and must be manually recovered later, to
+  avoid repeatedly retrying in the event of failure.
+
+  Returns `true` if a message was sent successfully, `false` if it failed, and
+  `nil` to indicate there was no sendable message in the database."
+  [ds f now]
+  (when-let [message
+             (jdbc/with-transaction [tx ds]
+               (when-let [message (some-> (sql/select-message-for-sending tx now)
+                               (db->scheduled-message))]
```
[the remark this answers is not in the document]
gws: There is, but I'm going to revert this based on a discussion with Mike anyway and will make sure the indentation sucks less.

src/scheduled_message_works/sending.clj
gws: Right, it's just supposed to be the intent of the query to mark a message as sent. `mark-sent` and `mark-failed` are probably better choices.

src/scheduled_message_works/db.clj
```diff
@@ -0,0 +1,212 @@
...
+  semantics."
+  [ds m]
+  (let [m' (scheduled-message->db m)]
+    (jdbc/with-transaction [tx ds]
+      (sql/delete-message-by-strict-identifiers tx (:message/identifiers m'))
+      (let [ret (merge (sql/insert-message! tx m')
+                       (sql/insert-message-amqp! tx m'))]
+        (db->scheduled-message ret)))))
+
+(defn run-with-sendable-message
```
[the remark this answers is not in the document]
gws: I'll see if I can come up with something better!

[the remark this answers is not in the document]
eggsyntax: Actually if `f` isn't _necessarily_ going to send the message, maybe this fn shouldn't mark it as sent (I'm thinking here of the part of the docstring that says "`f` is a function that is expected to have side effects, *likely* at least sending a message to an external system" (emp mine)). I guess the question is whether we expect `run-with-sendable-message` to ever be used by anything besides the `start-sender` loop.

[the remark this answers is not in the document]
gws: It's possible that the rename of `sql/send-message` will clear this up - the point here is to avoid exposing specifics of the database transaction to this part of the code - here, you only need to supply a function that performs side effects, and the bookkeeping of marking a message sent or failed (which is a transactional thing necessarily) is in `db.clj`. I'm still going to try for a better name.

src/scheduled_message_works/handlers.clj
```diff
@@ -1,144 +1,142 @@
...
+    (if (s/invalid? conformed)
+      (spec-error ::schedule-message req)
+      (let [msg (db/schedule-message ds conformed)]
+        (log/infof "Scheduled message ID %s with identifiers %s"
+                   (:id msg)
+                   (:identifiers msg))
+        {:status :ok
+         :operation :upsert
+         :message msg}))))
+
```
eggsyntax: Maybe worth creating a little helper for that conform-and-error-if-invalid pattern, since that shows up a few times.

src/scheduled_message_works/sending.clj
```diff
@@ -0,0 +1,76 @@
...
+              (recur)))))
+      (log/info "Stopped message sender"))
+    shutdown-fn))
+
+(defstate sender
+  :start (let [sleep-ms (get-in config [:sending :period-ms])]
+           (if (get-in config [:sending :enabled?])
+             (do
+               (log/info "Starting message sender")
+               {:shutdown (start-sender data-source
```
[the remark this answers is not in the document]
gws: I don't know if I addressed your concerns in my later changes, but when stopping the sender, I use a `let` binding to clarify what function we're calling to shut it down.

src/scheduled_message_works/db.clj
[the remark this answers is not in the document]
gws: I rewrote the docs and renamed this to clear up that this is really supposed to be sending messages, and the existing sending function in `sending` is an AMQP-specific sending function (I wrote this component keeping in mind that one day we may choose to send messages out over other media such as HTTP or Kafka, but given the nature of our work, we likely aren't going to do away with this kind of component altogether in the near future).

resources/sql/001-initial.up.sql
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+-- message
+
+CREATE TABLE IF NOT EXISTS "message" (
+  "id" uuid PRIMARY KEY,
+  "send_after" timestamptz NOT NULL,
+  "send_before" timestamptz,
+  "content" text NOT NULL,
+  "identifiers" jsonb NOT NULL,
```
dzcode: Is there a performance advantage/disadvantage of storing this as a `jsonb` vs a more relational foreign key with another table?

resources/sql/001-initial.up.sql
```diff
@@ -0,0 +1,78 @@
...
+);
+
+CREATE INDEX IF NOT EXISTS "message_send_after_idx"
+  ON "message" ("send_after");
+CREATE INDEX IF NOT EXISTS "message_send_before_idx"
+  ON "message" ("send_before");
+CREATE INDEX IF NOT EXISTS "message_identifiers_idx"
+  ON "message" USING gin ("identifiers" jsonb_path_ops);
+CREATE UNIQUE INDEX IF NOT EXISTS "message_identifiers_unique_idx"
+  ON "message" ("identifiers")
```
dzcode: Is there a reason this isn't also a `jsonb_path_ops` index?

resources/sql/001-initial.up.sql
gws: Yeah, definitely.

1. Identifiers are chosen arbitrarily by services, and aren't known up front, and would require either an EAV-style table or dynamically creating and keeping track of the identifier tables
2. It's desirable to enforce uniqueness across a given set of identifiers, which is more difficult to do

The `jsonb` column design ends up being much simpler to implement and understand, and the indexing using `jsonb_path_ops` is fast and well-suited for key/value lookups.

src/scheduled_message_works/db/sql.clj
```diff
@@ -0,0 +1,145 @@
...
+  messages."
+  [connectable now]
+  (jdbc/execute-one! connectable
+   ["SELECT m.id, m.send_after, m.send_before, m.content, m.identifiers,
+            ma.routing_key
+     FROM message AS m
+     JOIN message_amqp AS ma
+     ON m.id = ma.message_id
+     WHERE m.send_after < ?
+       AND (m.send_before > ? OR m.send_before IS NULL)
```
dzcode: I _think_ there should also be indices on `(send_before, state=scheduled)` and `(send_after, state=scheduled)` to get the performance gains of both indices, since otherwise I think it'll just choose the better one and ignore the other. This could also be added later if needed though

resources/sql/001-initial.up.sql
gws: Yeah, because we're enforcing uniqueness over the whole `jsonb` value, so we don't need to index the keys and values separately which is what `jsonb_path_ops` is designed for.

dzcode:
This looks good to me overall, though I mostly focused on the SQL parts since that's where I'm most familiar. Added a few comments around indexing and performance thoughts, but has been a while so take with a grain of salt

resources/sql/001-initial.up.sql
dzcode: Got it. In that case, similar to my comment below (and if my logic below is right), I think this index should include `state=scheduled` as well to get the performance gains (or have another index that has that included)

src/scheduled_message_works/db/sql.clj
gws: This is a good point, I wrote this originally before I decided to keep sent messages. We might start to see this query slow down over time without it assuming sent messages aren't cleared out regularly (right now we're keeping them in case we need stats). To avoid index bloat I might switch over to *just* using those partial indexes since `state = 'scheduled'` are the only states for which we care about performance.

src/scheduled_message_works/sending.clj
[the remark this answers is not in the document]
eggsyntax: WFM!

src/scheduled_message_works/db.clj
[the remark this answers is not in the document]
eggsyntax: :+1: 
I think especially the change to `mark-message-sent`/`mark-message-failed` makes things much clearer. Thanks!

resources/sql/001-initial.up.sql
gws: In that specific case I think the extra index work might not be worth it, since identifiers are unique across all scheduled messages anyway (enforced for scheduled ones) and we're clearing them out after ~a few months, it might cause us to look at a handful of rows once in a while but otherwise shouldn't degrade performance enough to justify the extra indexing.

However, I do think the send_after / send_before do need the extra state index because old messages are always going to be the first ones scanned in the `send_after` case, so if we're not deleting sent messages we'll always have to scan through old messages before finding eligible messages.

PR #87, opened 2019-09-19
Add optional :limit to search-messages

mrichards42:
More often than not, when I'm searching scheduled messages, all I really care about is if _any_ messages exist for the given set of identifiers (e.g. "are there any notifications for this election"), and I don't need to get the whole set back.


src/scheduled_message_works/db/sql.clj
```diff
@@ -17,15 +17,16 @@
...
    ["SELECT m.id, m.send_after, m.send_before, m.content, m.identifiers,
             ma.routing_key
      FROM message AS m
      JOIN message_amqp AS ma
      ON m.id = ma.message_id
-     WHERE m.identifiers @> ?::jsonb AND m.state = 'scheduled'"
-    identifiers]))
+     WHERE m.identifiers @> ?::jsonb AND m.state = 'scheduled'
+     LIMIT ?"
+    identifiers (or limit "ALL")]))
```
mrichards42: Ah, so this won't work -- standby

src/scheduled_message_works/specs.clj
```diff
@@ -10,3 +10,5 @@
 (s/def :message/identifiers
   (s/and (s/map-of keyword? (complement coll?))
          (complement empty?)))
+
+(s/def :sql/limit (s/nilable (s/and number? pos?)))
```
gws: Could this be `pos-int?`?

mrichards42: Learn something new every day -- changed!

PR #89, opened 2019-09-29
Performance improvements for high-volume sending

gws:
This PR addresses the following problems:

1. Sending a high volume of messages with a high number of scheduled messages results in rapid slowing of the rate at which messages can be sent
2. The PostgreSQL query optimizer switches away from using an index based on `send_after` once the table has a large number of messages marked `sent`
3. We can rapidly consume all available database capacity without some form of rate limiting

with the following corresponding solutions:

1. **Do more aggressive autovacuuming on the `message` table**. PostgreSQL MVCC requires old tuples to hang around in case they are visible by any transaction, and the autovacuum daemon comes around to remove them some time later. In our case, we are reading exactly the same index pointing to the same tuples that are rapidly being updated, which results in a dramatic slowdown without more aggressive removal of the dead tuples - the default of something like 10% of all table tuples (rows) on a typical 10M row table we tend to see means autovacuuming *doesn't even kick in* until 1 million messages have been sent, by which time querying has slowed down far too much to be adequate for our high volume sending days. More typical access patterns would not see a performance hit like this because accesses are often spread out over different parts of the index.
2. **Be explicit about the order in which rows should be retrieved**. By forcing an ORDER BY **that can use an index**, we keep the optimizer from making bad decisions about which index to traverse as the number of sent rows in the table grows. Those rows all have a `send_after` that match the message sending predicate `send_after < NOW()`, so PostgreSQL starts to believe the index is useless compared to a table scan. Forcing an ORDER here causes PostgreSQL to always use the index we want, and makes explicit the order we want to send messages in.
3. **Rate limiting has been added** and tested on a production workload. Without some form of rate limiting, message sending can consume database resources without bound. Having a rate limit in place allows us to decide exactly how fast we would like to send messages while consuming a predictable amount of database resources.


resources/sql/001-initial.up.sql
```diff
@@ -25,6 +25,23 @@ CREATE TABLE IF NOT EXISTS "message" (
...
+  -- therefore require a read attempt by every sending query (which all read the
+  -- index in the same order), quickly causing a lot more I/O work per read
+  -- for the database and therefore causing message sending to slow down
+  -- significantly on large tables after a short period of time (usually
+  -- single-digit minutes) without more aggressive vacuuming.
+  --
+  -- Vacuuming runs if the following inequality holds:
+  -- dead_tuples > threshold + (scale_factor * rows)
+  autovacuum_vacuum_scale_factor = 0.0,
+  autovacuum_vacuum_threshold = 1000
```
mrichards42: Can this change still go in the original migration file, or does this need to be 002?

resources/config.edn
```diff
@@ -80,4 +80,5 @@
  :sending {:enabled? #boolean #or [#env "SENDING_ENABLED" "true"]
            :batch-size #long #or [#env "SENDING_BATCH_SIZE" "10"]
            :concurrency #long #or [#env "SENDING_CONCURRENCY" "10"]
-           :period-ms #long #or [#env "SENDING_PERIOD_MS" "30000"]}}
+           :period-ms #long #or [#env "SENDING_PERIOD_MS" "30000"]
+           :rate-limit-s #long #or [#env "SENDING_RATE_LIMIT_S" "500"]}}
```
mrichards42: Looking back at #80 it looks like at that point we were able to send about 25/s max, so this is a jump of more than an order of magnitude. First of all, 👏 amazing work! Secondly, do we need to send messages that fast? I'm not sure we can move them through the rest of our system at more than like 100/s

src/scheduled_message_works/db/sql.clj
```diff
@@ -49,6 +49,7 @@
      WHERE m.send_after < ?
        AND (m.send_before > ? OR m.send_before IS NULL)
        AND m.state = 'scheduled'
+     ORDER BY m.send_after
```
eggsyntax: I'd love to see a comment here (or a pointer to one) explaining why the ORDER BY clause is important (maybe just a copy/paste of point 2 in your PR description).

resources/sql/001-initial.up.sql
gws: Normally, yes! But I am applying/testing these migrations before committing them here so... not this time :see_no_evil:

resources/config.edn
gws: Yeah, so right now our SES upper limit is 700/s and our Twilio upper limit is 500/s (segments, I think, so a much lower s-m-w throughput is adequate) - I'm thinking something like 900/s would be plenty, but 750/s might also be a nice number? What's in this PR is just a fast-enough default, so we can and should adjust this specifically for production!

I was seeing messages get sent every 1-3ms, and depending on how backed up the DB is, the query to pull them can happen in anywhere from 1-1000+ms. If everything is humming along smoothly, each thread can drive about 250 - 400 messages/sec (2ms per message * 100 messages) + (5ms query * 10 queries) = 250ms for 100 messages in the best case. Doing queries in parallel can help amortize the cost of queries that are taking more time than usual (e.g. if vacuuming is falling behind), but I think we don't need 10 threads doing sending - 5 should be plenty.

gws: I didn't say this explicitly but I think you and I would both rather see a s-m-w sending rate that's high enough to build up a bit of a Twilio backlog but not full-bore 3k messages/sec. I don't think RabbitMQ loves to have really long queues anyway - the fastest queue will always be an empty one (not that RMQ message rate is super important considering how relatively slowly we send emails/SMSes).

gws:
@mrichards42 I've updated the defaults based on some things I've learned recently. We can always add another replica if things are too sluggish!

resources/config.edn
mrichards42: Yep, absolutely agreed. I think the ultimate goal is to size things so that all our consumers are kept busy with work, and we aren't building up an unreasonable twilio backlog. I honestly have no clue what that looks like -- I think we were able to build up both an SES and a twilio backlog last year at certain points, though not throughout the day (since we stagger emails an sms).

750/s sounds totally reasonable to me, and not a big deal if we need to spin up another replica to speed things along.

gws: :100:, and the goal I have today for `email-works` and `sms-works` is to set them up to be probably _under_ what we'll want on a large sending day in the default 3-replica configuration, but to where we can pretty easily adjust replicas upward by 1-3 or so to get the actual sending rates we want. Same with this service - I mainly don't want us to have to make commits to these service repos in real time order to adjust rates up or down.

PR #92, opened 2020-01-22
VE-130 Reschedule handler

mrichards42:
Basically a copy of the `schedule` handler, but using strict update semantics. I also chose to make `:message` optional, in which case the existing content is preserved.


gws:
@mrichards42 This is similar to something I [worked on on Monday](https://github.com/democracyworks/scheduled-message-works/tree/add-smw-update-handler) but you beat me to getting it shipped. Overall it looks good though, and if we want to switch to using this soon we should probably get it merged.

mrichards42:
Ah! Sorry for the duplication. I was starting to get a little nervous about time and didn't realize you were already on it. Either way is fine by me.

src/scheduled_message_works/db.clj
```diff
@@ -112,6 +112,11 @@
       (update :created-at #(or % (Instant/now time/*clock*)))
       (set/rename-keys m))))
 
+(defn- scheduled-message-with-optional-content->db
+  [msg]
+  (cond-> (scheduled-message->db msg)
+    (not (find msg :message)) (dissoc :message/content)))
```
gws: It feels to me like this should compose with `scheduled-message->db` instead of being a function that calls it, so you'd do something like

```clojure
(-> msg (scheduled-message->db) (dissoc-optional-content))
```

down below

gws:
@mrichards42 no need to apologize, I didn't say anything about it!

src/scheduled_message_works/db.clj
mrichards42: I like that plan, but if you take a blank message and run it through `scheduled-message->db` you end up with
```clj
{:message/id #uuid "02cf27be-1ddf-4648-a039-b6220f3c970d"
 :message/content "nil"
 :message/identifiers "{}"
 :message/created_at #object[java.time.Instant "0x3bfba6f" "2020-01-22T23:06:23.369Z"]}

;; and we could write this, but it feels wrong
(defn dissoc-optional-content [m]
  (cond-> m
    (= "nil" (:message/content m)) (dissoc :message/content)))
```

src/scheduled_message_works/db/sql.clj
```diff
@@ -87,6 +87,35 @@
...
+      (:message/send_after m) (:message/send_before m) (:message/content m)
+      (:message/identifiers m)])
+    (jdbc/execute-one! connectable
+     ["UPDATE message
+      SET send_after = ?, send_before = ?
+      WHERE state = 'scheduled' AND identifiers = ?::jsonb
+      RETURNING id, send_after, send_before, content, identifiers,
+                state, created_at"
+      (:message/send_after m) (:message/send_before m)
+      (:message/identifiers m)])))
```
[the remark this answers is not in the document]
mrichards42: *sigh* yeah -- I almost rolled that by hand, but it was the same number of lines and way more confusing. We could _almost_ do
```clj
(next.jdbc.sql/update!
 connectable
 "message"
 {:send_after ...
  :send_before ...
  :content ...}
 {:identifiers ...})
```
but I don't think you can get it to handle the `jsonb` cast, and I don't think you can get `RETURNING` either.

resources/config.edn
```diff
@@ -20,6 +20,16 @@
                        :response nil
                        :prefetch-limit #read-edn #or [#env "SCHEDULE_MESSAGE_PREFETCH" "25"]
                        :threads #read-edn #or [#env "SCHEDULE_MESSAGE_THREADS" "50"]}
+                      {:queue "scheduled-message-works.message.reschedule"
```
mrichards42: Any concerns about naming? Looks like you were going for `message.update` rather than `message.reschedule`.

src/scheduled_message_works/db.clj
gws: Oh yeah the function would have to take two arguments to get at what the original is doing, I don't feel incredibly strongly about it though so let's just leave it

resources/config.edn
gws: It's true, but I was also tackling by-id and by-identifiers, this behavior is more narrowly scoped to "rescheduling" so it seems like a fine name for the operation that's being done.

PR #93, opened 2020-02-01
Split sender into a singleton deployment

gws:
Not pictured are the HPA settings which call for 6 replicas.

First, we allow the connection pool to grow up to 50 connections. The reason this was initially restricted was to try to mitigate the scaling issues, which we have since found are unrelated to connection creation during scaling events. This makes the service more responsive to load increases, but keeps a lighter touch on the backend when things are quiet. Having a few warm connections around helps a bit when responding to bursts of activity.

The memory request/limit can also come down from 2Gi to 1Gi. We typically use between 300-500Mi, even at full capacity, so 1Gi is a pretty healthy margin. This is a normal heap size for JVM-based non-Datomic services.

Finally, we've created a separate deployment for just the sending so that it's easier to adjust the rate limit and allow for free autoscaling in the future.


gws:
It should be noted that we've been running with these settings in production through the Facebook pushes this week, I had to adjust the sending rate limits on the fly as we scaled up the replicas to avoid overwhelming the rest of our systems with sending. With the lighter memory usage, maybe we can just leave the replica count at 6 all year 'round and avoid fiddling.

deploy/tv/production/scheduled-message-works.configmap.yml
```diff
@@ -6,8 +6,8 @@ metadata:
   namespace: production
 data:
   HIKARICP_MINIMUM_IDLE: "25"
-  HIKARICP_MAXIMUM_POOL_SIZE: "25"
+  HIKARICP_MAXIMUM_POOL_SIZE: "50"
   JDBC_HOST: "tv-scheduled-message-works-production.cluster-c3tl1picvd92.us-west-2.rds.amazonaws.com"
-  SENDING_RATE_LIMIT_S: "125"
+  SENDING_RATE_LIMIT_S: "63"
```
dzcode: Is this always a direct function of the `HIKARICP_MAXIMUM_POOL_SIZE`? Not for this PR, but would there be a way to hard-code
`MAX_RATE_LIMIT_ACROSS_REPLICAS` and `NUMBER_OF_REPLICAS` to deduce this number rather than hiding the math behind a magic number?

gws: No, it's a function of replica count and how many messages we want to send per second in order to avoid stressing the system too much.

A better way to solve this would be to introduce a per-service rate limiter, or to use a template system such as Helm or Kustomize to do the `desired_send_rate / num_replicas` math.

mrichards42: Would it be worth splitting this service into schedulers (no sending) and a single sender (no scheduling) so we could directly control the sending rate? Almost like we did before the postgres rewrite. Not saying that's necessarily an optimal solution, but scaling and config for scheduling and sending are more or less independent.

gws: I don't think we need to restrict the scheduling but yeah I like this compromise, I just pushed a commit that creates a singleton Deployment (with the Recreate deployment strategy) with the new rate limit, will re-request and merge on a :+1:!

deploy/tv/scheduled-message-sender.deployment.yml
```diff
@@ -0,0 +1,72 @@
...
+      annotations:
+        ad.datadoghq.com/scheduled-message-works.check_names: |
+          ["prometheus"]
+        ad.datadoghq.com/scheduled-message-works.init_configs: |
+          [{}]
+        ad.datadoghq.com/scheduled-message-works.instances: |
+          [
+            {
+              "prometheus_url": "http://%%host%%:9090/metrics",
+              "namespace": "scheduled_message_works",
```
gws: I still want to group these by `scheduled-message-works` even if one of the pods behaves a little differently.

deploy/tv/scheduled-message-sender.deployment.yml
```diff
@@ -0,0 +1,72 @@
...
+            memory: 1Gi
+          requests:
+            memory: 1Gi
+        envFrom:
+        - configMapRef:
+            name: scheduled-message-sender
+        - secretRef:
+            name: turbovote-common
+        - secretRef:
+            name: scheduled-message-works
```
gws: Also OK - secrets should still change in one place.

deploy/tv/production/scheduled-message-sender.configmap.yml
```diff
@@ -0,0 +1,13 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: scheduled-message-sender
+  namespace: production
+data:
+  HIKARICP_MINIMUM_IDLE: "25"
+  HIKARICP_MAXIMUM_POOL_SIZE: "50"
+  JDBC_HOST: "tv-scheduled-message-works-production.cluster-c3tl1picvd92.us-west-2.rds.amazonaws.com"
+  SENDING_RATE_LIMIT_S: "350"
```
mrichards42: The idea is that we'd only run 1 pod? Should we be worried about redundancy if this pod hangs? Or is that going to be straightforward to catch?

gws: If the pod did hang (we haven't seen that yet with s-m-w pods) we'd at some point be in the same situation having more than 1, for better or worse, since there are other replicas around serving traffic. To tell you the truth I'm not sure why we run 3 of everything other than for historical reasons. This mainly makes it as easy as possible to configure the rate limit + sending thread count, I'm open to running more than 1 replica though.

gws: We could set up a heartbeat mechanism inside s-m-w that ticks whenever the sending thread loops around to check for messages and alert on a lack of it, too.

mrichards42: I think fairwinds has some alerts set up around deployments with 0 pods actually, so I'm not super worried.

gws: I like that idea, and I haven't written any code in, like, a week, so maybe I'll tackle that!

gws: @mrichards42 Unfortunately like with `election-works` we wouldn't get notified using those alerts since Kubernetes thought they were alive. We don't have health checks implemented for our services but if we had one for `election-works` then Kubernetes would have recreated those pods back in January and we may never have noticed the issue.

mrichards42: Ah good point -- health checks would be a good move

PR #94, opened 2020-02-17
Add channel for check-your-registration sending

dzcode:
Companion to [this PR](https://github.com/democracyworks/election-notification-works/pull/238)


mrichards42:
Look good, but you also will want to add that to `resources/config.edn`. Looks like we always configure these channels as `:outgoing-events`, which is a little weird, but since we aren't expecting a response it shouldn't matter too much.

PR #97, opened 2020-04-16
[WIP] Reporting query

mrichards42:
This is a "master" reporting query for scheduled messages. The idea is to expose this in tv admin, but it could also be used for one-off chores. For instance, I added `:limit` to message search a while back b/c I wanted to be able to know if all election notifications were deleted. Now, a better way to do that might be:

```clj
;; message to scheduled-message-works.message.summarize
{:filters {:identifiers {:election-id #uuid ""}}}
;; returning
{:status :ok, :summary [{:election-id #uuid "", :count 100}]}
;; or
{:status :ok, :summary []}
```

There are no tests yet, mostly I'm looking for feedback on the interface, which looks like:

```clj
{:groups [:identifiers/election-id, :identifiers/medium]
 :resolution :day
 :filters {:identifiers {:election-id :*
                         :notification-type :upcoming-election}
           :after #inst "2020-05-01"
           :before #inst "2020-06-01"}}
```

producing the following query

```sql
SELECT
  date_trunc('day', send_after) AS send_bucket,
  identifiers->'election-id' AS election_id
  identifiers->'medium' AS medium,
  COUNT(*)
FROM
  message
WHERE
  state='scheduled'
  AND identifiers @> '{"notification-type": ":upcoming-election"}'
  AND identifiers ?& arrray['election-id']
  AND send_after >= '2020-05-01'
  AND send_after <= '2020-06-01'
GROUP BY
  send_bucket,
  election_id,
  medium
ORDER BY
  send_bucket
```


mrichards42:
> Are :groups also meant to allow just identifiers? I'm wondering if it's possible to have both that and the :identifiers under :filters accept namespaced keywords (or not) consistently.

`:groups` was originally supposed to only allow un-namespaced identifiers, but it occurred to me that grouping by `:send-to` might be a useful option at some point. I didn't feel like writing out the join code to actually support that right now, but I decided to throw a namespace on `:groups` to allow that possibility.

Now that you mention it though, I wouldn't mind making filters a flat map like:
```clj
{:identifiers/election-id :*
 :identifiers/notification-type :upcoming-election
 :send-after {:after #inst "2020-05-01", :before #inst "2020-06-01"}}
;; which also allows potentially
{:send-before {:before #inst "2020-07-01"}}
```

gws:
Nice, that'd be good too.

test/scheduled_message_works/summary_test.clj
```diff
@@ -0,0 +1,176 @@
...
+           (set (->summary {:resolution :second}))))
+    ;; TODO: there are some time zome issues w/ buckets >= day
+    #_(testing "day resolution"
+        (is (= #{{:count (count-for :april) :send-bucket #inst "2020-04-01"}
+                 {:count (count-for :may) :send-bucket #inst "2020-05-12"}}
+               (set (->summary {:resolution :day})))))
+    #_(testing "month resolution"
+        (is (= #{{:count (count-for :april) :send-bucket #inst "2020-04"}
+                 {:count (count-for :may) :send-bucket #inst "2020-05"}}
+               (set (->summary {:resolution :month})))))))
```
mrichards42, line 126: If you have any thoughts about how to fix this issue I'd love to hear them. What I'm seeing is that these timestamps are returned offset by the jvm's time zone, so:

from a psql shell
```sql
SELECT date_trunc('day', NOW())

-- 2020-04-22 00:00:00+00
```

in scheduled-message-works
```clj
(jdbc/execute-one! data-source ["SELECT date_trunc('day', NOW())"])
; => {:date_trunc #object[java.time.Instant "0x27413cb7" "2020-04-22T04:00:00Z"]}
```

This is likely fine in prod, since the local time zone for both servers is likely UTC, but it makes testing locally impossible.

src/scheduled_message_works/db.clj
```diff
@@ -265,3 +272,59 @@
...
+                               (map (comp name key)))
+        identifiers-contain (some->> filters
+                                     (medley/filter-keys sql/identifier-key?)
+                                     (medley/remove-vals wildcard-val?)
+                                     (medley/map-keys name)
+                                     (not-empty) ;; don't generate "{}"
+                                     (identifiers->db))
+        groups (->> groups
+                    (map-indexed (fn [i kw] [kw (keyword (str "group_" i))]))
+                    (into {}))]
```
mrichards42, line 294: Not a change in the interface, but this should plug up a hole where if you passed a group like `:identifiers/you'broke'my'query`, the generated sql would have been:

```clj
["SELECT identifiers->? AS you'broke'my'query" "you'broke'my'query"]
```

which obviously is bad. Now it's

```clj
["SELECT identifiers->? AS group_0" "you'broke'my'query"]
```

test/scheduled_message_works/summary_test.clj
gws: OK I spent far too long on this, but the gist is that `timestamp` has no associated timezone and is in the SQL standard for compatibility. However, `date_trunc` deals in it, I think due to the nature of the function, chopping times off isn't very time-mathy. `timestamp` isn't otherwise recommended, it's recommended to use the (confusingly named) `timestamptz` or `timestamp with time zone` which I say is confusing because it actually converts anything you give it into UTC first, then stores it that way *without* time zone information (that's the confusing part). However at least then PostgreSQL can assume it's UTC and is free to convert at will on output based on connection time zones, etc without making incorrect assumptions.

We store all of our timestamps using `timestamptz`, so we know our data has been converted and stored correctly, at least.

However, to get the results you want, you have to do two things, **assuming you want your resolution math to use UTC**:

1. Shift the input time into UTC explicitly before feeding it to `date_trunc` which is going to just do math with the numbers and do no zone conversion, and
2. Stick UTC back onto the output of `date_trunc` since that output will be otherwise zoneless. This will allow any conversion due to your connection time zone to be done correctly rather than the output you got.

Assuming I'm not way off here, try this:

```sql
SELECT timezone('UTC', date_trunc('day', timezone('UTC', NOW())));
```

`NOW()` may be the current time in your time zone, that's OK. It first gets shifted into UTC before the time zone is promptly dropped by `date_trunc`. Then the "zoneless" result gets `UTC` added back on at the end so your JVM/connection conversion doesn't show the incorrect point on the timeline like you were seeing above. For me, for instance, something like:

```sql
postgres=# SELECT timezone('UTC', date_trunc('day', timezone('UTC', '2020-04-22T18:30:00-0700')));
        timezone
------------------------
 2020-04-22 17:00:00-07
(1 row)
```

is correct, because that's the start of the **23rd** of April in UTC. When outputting that as UTC, you'd get midnight on 2020-04-23.

[`timezone` is documented](https://www.postgresql.org/docs/10/functions-datetime.html#FUNCTIONS-DATETIME-ZONECONVERT) to do different things based on the datatypes it's handed, and might be useful to review. You can also write it `x AT TIME ZONE 'UTC'` if you'd rather not use the function form - they're the same thing.

And finally, if you want to use Eastern time for your groupings, you'd use `America/New_York` instead of `UTC` above. As long as it's done in both places, the output should make sense.

gws: It really seems kind of confusing and unnecessary until you stop and think about what `date_trunc` is actually doing - it's really a hatchet job on numbers rather than doing anything meaningful with DST or time zones, so doing this conversion on the in and out truly does make sense. Or maybe that's just the Stockholm syndrome talking.

mrichards42: I hacked at this a while back when, but didn't catch that I needed _two_ timestamps. I think I finally see what's going on.

&lt;rant>
I don't know what to say but WAT

> timestamp without time zone AT TIME ZONE zone => timestamp with time zone
> timestamp with time zone AT TIME ZONE zone => timestamp without time zone

I get why you need both these conversions, but to call them the same thing is ridiculous.
&lt;/rant>

OK, so . . .

```sql
timezone('zone', send_after)    -- remove the timezone (adjusting the timestamp into this time zone)
date_trunc('day' ^^^)           -- truncate the timestamp
timezone('zone', ^^^)           -- add the timezone back

-- which means
postgres=> SELECT timezone('America/New_York', date_trunc('day', timezone('America/New_York', '2020-04-01T00:00:00Z'::timestamptz)));
        timezone
------------------------
 2020-03-31 04:00:00+00
(1 row)

-- and also
postgres=> SELECT timezone('America/New_York', date_trunc('day', timezone('America/New_York', '2020-04-01T00:00:00-04:00'::timestamptz)));
        timezone
------------------------
 2020-04-01 04:00:00+00
(1 row)
```

mrichards42: And I definitely agree with you that `date_trunc` can't possibly deal with time zones, since like you say, it's just doing simple math.

gws: Yeah all of that looks right to me... for what it's worth, considering I also learned half of this yesterday :joy:

project.clj
```diff
@@ -15,7 +15,9 @@
                  [democracyworks/kehaar "1.0.3"]
                  [diehard "0.9.2"]
                  [digest "1.4.9"]
+                 [honeysql "0.9.10"]
                  [io.prometheus/simpleclient_hotspot "0.8.1"]
+                 [medley "1.3.0"]
```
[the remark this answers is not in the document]
mrichards42: I've really grown to ❤️ medley recently. It's great code, and it often makes my code better too.

mrichards42:
@gws OK, after the whole `date_trunc` conversation, I went ahead and added `:timezone` to the message: https://github.com/democracyworks/scheduled-message-works/pull/97/files/8ddf5e74ff276d727b72cf5c8e64b0c203ef8fdb..c3375ede4ea3bb21224f445965b63f70da86ef19

gws:
@mrichards42 :shipit:

:clock3: :clock4: :clock5: :clock6: :clock7: :clock8:

.clj-kondo/config.edn
```diff
@@ -0,0 +1 @@
+{:lint-as {mount.core/defstate clojure.core/def}}
```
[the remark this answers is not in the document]
mrichards42: I'd love to at some point. Mostly just doing this to make dev a little less noisy for now.

